PR #234, opened 2022-08-25
Enable Custom Faces

alxjrvs:
This PR will enable `Custom Faces` on your rolls. 

It is only supported in the "Randsum Options" call, and replaces the 'sides' option. When you define faces, give the key an array of sides. those sides will represent the faces of the die, informing the roller how many sides it has implicitly. 

You can add strings, numbers, mix and match - whatever your poison.


alxjrvs:
While this approach *works*, it also means not using the optimal UX - just passing in an array for `sides` rather than a number - and so I will be retiring this approach. However, this basic model - keeping `faces` as an internal array - will stick around.

PR #258, opened 2022-08-28
Condense files, compare sizes

alxjrvs:


alxjrvs:
Avoiding this level of change, going with a milder approach

PR #301, opened 2022-10-08
use DTS-CLI

alxjrvs:
Though I learned a great deal doing this my own way, I found a helpful CLI toolkit that I have decided to hitch my particular wagon to. 

Long story short, I am curious what - if any - result this will have on my bundle size, given the community opinions vs. my own home-rolled versions.


alxjrvs:
Merging this because the `size` workflow, when setup this way, will always break (as it tries to compare the base branch - which does not have `size-limit` - with this one). 

I could:
 - remove the workflow from this PR
 - merge this PR 
 - open a new PR with just the workflow

But!

Issue #995, opened 2026-03-12
fix(codegen): use input description as validation label; add string enum validation


alxjrvs:
### 1. Validation label uses spec name instead of input description
Codegen generates `"D&D 5th Edition modifier"` but fifth tests expect `"5E modifier"`. Fix: when `InputDeclaration.description` is present, use it as the validation label. Fallback: `"${spec.name} ${fieldName}"`.

### 2. ValidationError prefix in error messages
Codegen produces `'Validation error: D&D 5th Edition modifier must be...'` but tests expect bare `'5E modifier must be...'`. The `validateFinite`/`validateRange` calls should throw plain `Error`, not wrap in `ValidationError`. Check whether the codegen is generating calls that go through a wrapper class.

Root-rpg tests pass (19/19) because they likely assert `.toThrow()` without checking the message. Fifth tests assert exact messages — hence the failures.

### 3. String enum validation not generated
`roll({ rollingWith: 'invalid' })` should throw but doesn't. When a string input has `enum`, codegen must emit a guard:
```ts
if (input.rollingWith !== undefined && !['Advantage', 'Disadvantage'].includes(input.rollingWith)) {
  throw new Error(`Invalid rollingWith: ${input.rollingWith}. Must be one of: Advantage, Disadvantage`)
}
```

## Fixes

1. `buildInputType` / validation codegen: use `decl.description ?? \`${spec.name} ${fieldName}\`` as label
2. Verify generated validation calls use raw `validateFinite`/`validateRange` without `ValidationError` wrapper
3. Add enum guard generation for string inputs with `enum` constraint

## Spec change needed (fifth.randsum.json)
Add `"description": "5E modifier"` to the modifier input so the generated label matches test expectations.

alxjrvs:
## Exact Validation Error Messages Expected by Tests

### Fifth (`games/fifth/__tests__/roll.test.ts` lines 95-127)

Tests use **exact string matching** via `.toThrow(string)`:

| Test | Input | Expected Error Message |
|------|-------|----------------------|
| NaN modifier | `{ modifier: NaN }` | `5E modifier must be a finite number, received: NaN` |
| Infinity modifier | `{ modifier: Infinity }` | `5E modifier must be a finite number, received: Infinity` |
| Modifier too high | `{ modifier: 31 }` | `5E modifier must be between -30 and 30, received: 31` |
| Modifier too low | `{ modifier: -31 }` | `5E modifier must be between -30 and 30, received: -31` |

[...]

Issue #940, opened 2026-03-11
[E-FATE] Epic: Fate Core Game Package


alxjrvs:
## Summary

Fate Core game package — Fudge dice (4dF) with Fate ladder labels. Implemented as a `.randsum.json` spec with codegen, following the same pattern as all other game packages in `@randsum/games`.

## Score

| Utility | Need | Impact | Feasibility | Mission | Total |
|---------|------|--------|-------------|---------|-------|
| 3 | 2 | 2 | 5 | 4 | **16/25** |

**Verdict**: Keep — bundle with a game-package sprint, do not prioritize in isolation

## Why

Genuinely trivial to implement — Fudge dice are `[-1, 0, 1]` and custom sides already work in the roller. Useful as a "look how easy this is" showcase. Good companion task for a game-package sprint alongside PF2e.

## Approach

> **Updated 2026-03-24**: Game packages now use `.randsum.json` specs + codegen, not the removed `createGameRoll` factory. See existing game packages (blades, fifth, pbta, etc.) in `packages/games/` for the pattern.

- Create `fate.randsum.json` spec defining 4dF with custom sides `[-1, 0, 1]`
- Codegen produces `src/fate.generated.ts` with `roll()` function
- Fate ladder labels (Terrible through Legendary) defined in the spec
- Accessible via `import { roll } from '@randsum/games/fate'`

## Features

- [ ] `fate.randsum.json` spec in `packages/games/`
- [ ] Fudge dice: 4dF producing -4 to +4 with Fate ladder labels (Terrible -> Legendary)
- [ ] Default 4 dice, configurable quantity
- [ ] Result includes numeric total + Fate ladder label
- [ ] Exported `FATE_LADDER` constant mapping values to labels
- [ ] Full test coverage with property-based tests
- [ ] Bundle within game package size limit

## Open Questions

- [ ] Should this include Fate Accelerated (same dice, different approach labels) or Fate Core only?
- [ ] Should the package handle Fate points / invoke bonuses, or keep it dice-only?

PR #1073, opened 2026-06-13
Remove the runtime spec interpreter (single-source spec semantics)

alxjrvs:
Stacked on #1072. Retires the `@randsum/games` runtime spec interpreter so spec semantics live in exactly one place — the code generator.

## Why
The interpreter (`loadSpec`/`loadSpecAsync` + `executePipeline`/`loader`/`modifierTranslator`/`conditionEvaluator`/`inputBinder`) was a **second, independent implementation** of spec semantics running alongside codegen — the monorepo's biggest structural duplication and a silent-drift risk (the dual-consumer test in #1072 caught two real interpreter bugs). It had **zero production consumers**; the public `@randsum/games/schema` exported it, but nothing imported it. Its main role was as the test harness for spec features.

## What
- **Deleted** the interpreter source (5 files) + its machinery unit tests.
- **New `compileSpec` test harness** — compiles a spec through `generateCode` and dynamically imports the result, so feature tests verify the **code we actually ship** rather than a parallel interpreter. Every future spec feature is now implemented and tested **once, not twice**.
- **Migrated** ~10 feature tests (degree-of-success, conditional/multi pools, post-resolve, details, dynamic sides, numeric results, multi-roll, …) to the harness. Two now assert codegen-*time* rejection (range gaps, missing-input enums) where the interpreter previously deferred to runtime — the generator catches these earlier.
- **BREAKING**: `loadSpec`/`loadSpecAsync` removed from `@randsum/games/schema`. Major bump via changeset. Use the generated subpath exports (e.g. `@randsum/games/blades`) or `generateCode`. If runtime spec-rolling is ever needed, build it as *generate-then-run* on the single codegen path.

## Validation
`bun run check` green across all 8 packages; gen:check, `bun audit --level=high`, and knip all pass. Games: 440 tests.

## Note on the other P2 items
- **Spec input `description`s** (audit finding): investigated and intentionally **not changed**. Per issue #995 the `description` field *is* the validation error label (opt-in, like fifth's `"5E modifier"`); the default `"{Game} {field}"` labels are already clear, and full-sentence descriptions degrade error messages. Low-value — left as-is.
- **Fate/custom die faces** (#940): deferred — needs a small design decision (codegen now single-path, so smaller) and TTRPG-accuracy review. Flagged for a follow-up.


alxjrvs:

PR #1074, opened 2026-06-13
Support custom die faces (unblocks Fate Core #940)

alxjrvs:
Stacked on #1073. Adds custom die faces to the `.randsum.json` spec format — the last marquee P2 item from the refresh audit. Single-path now that the runtime interpreter is gone (#1073).

## What
A pool may declare `faces` instead of `sides`:

- **Numeric faces** — `{ "faces": [-1, 0, 1] }` (a Fate/Fudge die) roll and sum as their values and feed outcome ranges. This unblocks Fate-style games (#940): `4dF` + an outcome ladder is now expressible.
- **String faces** — `{ "faces": ["hit", "miss", "crit"] }` (a "table die") with `resolve: "faces"` resolve to the rolled **label**, generating a face-label result union type (`'crit' | 'hit' | 'miss'`).

## How
- Roller's options object ignores a `faces` array; only notation produces custom faces. So codegen emits `<quantity>d{...}` notation for faces pools. Numeric faces use `total`; string faces read roller's `customResults` for the labels.
- Schema: `PoolDefinition` is now `sides` XOR `faces`; `resolve` gains `"faces"`.
- Range-coverage analysis computes the total range for numeric faces (sign-aware) and skips string/mixed faces.
- Faces are rejected with a clear `SchemaError` inside multi-pool (`dicePools`) / conditional pools (out of scope for now).
- All threaded through types → normalizer → codegen; **no existing game's generated output changes** (gen:check green). 444 games tests (incl. new Fate + table-die coverage).

## Validation
`bun run check` green across all 8 packages; gen:check, `bun audit --level=high`, knip all pass.

## Note
This is the *capability*. The actual `@randsum/fate` game package (#940) — with the full Fate ladder and accurate mechanics — is a follow-up that this unblocks, and would benefit from a TTRPG-accuracy review.


alxjrvs:

PR #1075, opened 2026-06-13
P2 backlog (dice-ui types, codegen escaping, @randsum/fate) + strict test-typecheck across all packages

alxjrvs:
Stacked on #1074. Two themes, all commits independently verified by re-running every check myself (not trusting agent self-reports — see Trust note).

## 1. P2 backlog
- **dice-ui / Expo native types**: deleted the hand-maintained `apps/expo/types/dice-ui.d.ts` shim; Expo typechecks against the real native barrel. Exported `DocModalProps`.
- **games codegen string escaping** (with test).
- **`@randsum/fate`** (#940), built on the custom-faces feature: 4dF + skill modifier + the Fate **ladder**. **Ladder corrected** to canonical (added Catastrophic −3 / Abysmal ≤−4 the agent folded into Terrible).

## 2. Strict test-typecheck (systemic gap you flagged)
Most packages' tsconfigs **excluded their test files from typechecking** — hiding `any`/type errors. Now every package typechecks its tests:
| Package | Was | Fixed |
|---|---|---|
| games `__tests__/lib` | excluded | **127 errors** fixed |
| dice-ui `.native.*` | whole RN target excluded | enabled; **1 real bug** fixed (native importing web props) |
| roller `*.test.ts` | only `.types.test.ts` checked | **208 errors** fixed; `traceRoll` param narrowed to the minimal interface it uses |
| cli, discord-bot | tests excluded | enabled (0 errors) |

No `any` / `@ts-ignore` / `as unknown as` introduced; no test assertions weakened (counts unchanged: roller 2719, games 478). `compileSpec` harness return type flattened so `details` is accessible in tests.

## Excluded / still open (honest)
- **`jarvis/p4-roller` from the backlog workflow was dropped** — that agent pushed a **hollow branch** (tests, no source; false "green" report). The roller *backlog* items (`sideEffects` real fix, regex DRY, `ARITHMETIC` coverage test, comma-AND docs) remain **not done**.
- **Fate** mechanics still want a human TTRPG-accuracy glance (modifier range is a design choice).

## Validation
`bun run check` (all 8 packages), knip, and `bun audit --level=high` all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1076, opened 2026-06-13
Roller P2 backlog: sideEffects publish-injection, die-marker DRY, ARITHMETIC coverage, comma-AND docs

alxjrvs:
Stacked on #1075. The four roller backlog items the workflow's roller agent **faked** (it pushed test files with no implementation and a false "all-green" report). Re-implemented properly this time, each verified.

## Items
1. **`sideEffects` real fix** (ADR-018). The in-repo `package.json` keeps `sideEffects: true` because (a) bunup's self-DCE breaks roller's `dist` when it builds with `false`, and (b) the cli bundles roller from source (`noExternal`) and reads that field. So instead, `scripts/publish.ts` injects `sideEffects: false` into the **published tarball** at pack time and restores immediately — `bun pm pack` doesn't rebuild dist, so consumers get the healthy built dist with a tree-shakeable manifest. **Verified**: tarball manifest `false`, dist healthy (docs 3.3k / trace 2.6k, not the broken stubs), in-repo restored to `true`, `node apps/cli/dist/index.js 2d20` exits 0.
2. **Die-type regex DRY**. The six special-die marker fragments were written 3× (isDiceNotation, tokenize, parseSpecialPool); isDiceNotation alone repeated them ~5×. Extracted canonical fragments to `notation/constants.ts`; isDiceNotation composes from them. New cross-check test exercises each marker through isDiceNotation + tokenize + `parseSpecialPoolSegment` so a future drift fails loudly. (+1B nudged `tokenize.js` over budget → bumped 6.5 → 6.75 KB.)
3. **`ARITHMETIC_MODIFIERS` coverage test**. Exported the trace map + a test asserting it stays exactly in sync with the registry's `Scale`-category modifiers.
4. **Comma-AND condition docs**. Documented that comparison operators combine via AND while bare numbers OR as exact matches — `{5,10}` is "exactly 5 or 10", **not** a range (use `{>=5,<=10}`).

## Validation
roller 2731 tests (+12 new), 0 fail; all 1111 notation-behavior tests unchanged. `bun run check` (all 8 packages), knip green.

This closes out the dropped backlog — nothing from the original audit remains open except the Fate TTRPG-accuracy glance (a human judgment call).

### DIFF
--- a/packages/roller/__tests__/notation/dieMarkers.crosscheck.test.ts
+++ b/packages/roller/__tests__/notation/dieMarkers.crosscheck.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, test } from 'bun:test'
+
+import {
+  DIE_MARKER_CUSTOM_FACES,
+  DIE_MARKER_DRAW,
+  DIE_MARKER_FATE,
+  DIE_MARKER_GEOMETRIC,
+  DIE_MARKER_PERCENTILE,
+  DIE_MARKER_ZERO_BIAS
+} from '../../src/notation/constants'
+import { isDiceNotation } from '../../src/notation/isDiceNotation'
+import { tokenize } from '../../src/notation/tokenize'
+import { parseSpecialPoolSegment } from '../../src/notation/parse/parseSpecialPool'
+
+/**
+ * Cross-check that the shared special-die marker fragments in constants.ts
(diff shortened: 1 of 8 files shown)


alxjrvs:

PR #1088, opened 2026-06-24
chore(release): publish via npm OIDC Trusted Publishing

alxjrvs:
## Why

The Release workflow 404'd publishing `@randsum/games` because the `NPM_TOKEN` secret (last set 2022-07-23) was a stale/expired token. npm now only issues **granular tokens with a 90-day max lifetime**, so any token-based `NPM_TOKEN` is a recurring expiry trap — it *will* break again on a timer.

OIDC **Trusted Publishing** removes the long-lived token entirely: GitHub Actions authenticates per-run via a short-lived OIDC id-token. The npm-side trusted publisher is already configured for `RANDSUM/randsum` → `publish.yml`, and the workflow already grants `id-token: write`.

## Changes

- **`publish.yml`** — drop `NPM_TOKEN` / `NODE_AUTH_TOKEN` from the Release job env. A token present in the env overrides OIDC, so it must be removed for npm to use the OIDC exchange. (`GITHUB_TOKEN` stays — changesets needs it.)
- **`scripts/publish.ts`** — add `--provenance`, guarded by `GITHUB_ACTIONS` so local `bun publish --otp` keeps working (provenance requires an OIDC-capable CI).
- Updated the security-invariant comment to reflect token → OIDC.

## Follow-up (npm side, manual)

- Confirm the trusted publisher is configured for **all three** published packages: `@randsum/roller`, `@randsum/games`, `@randsum/cli` (each package has its own publishing-access config). Screenshot confirmed at least one.
- The old `NPM_TOKEN` secret can be deleted once a real publish succeeds via OIDC.

## Validation

- `bun run build`, typecheck, lint, format all pass (pre-commit hooks green).
- OIDC publish can only be exercised by an actual version bump, so the first real publish post-merge is the live test. Token-skip behavior for already-published versions is unchanged.

### DIFF
--- a/.changeset/verify-oidc-publish.md
+++ b/.changeset/verify-oidc-publish.md
@@ -1,0 +1,5 @@
+---
+"@randsum/games": patch
+---
+
+Verify the npm OIDC Trusted Publishing release pipeline end-to-end. No functional or API change — release-infrastructure smoke test only.
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,8 @@ name: Release
 
 # SECURITY — workflow_run safe-usage invariant (audit L7 / security F3).
 # `workflow_run` runs in the PRIVILEGED base-repo context with access to
-# NPM_TOKEN / GITHUB_TOKEN and contents:write + id-token:write. To avoid a
+# GITHUB_TOKEN + the OIDC id-token (used for npm Trusted Publishing) and
+# contents:write + id-token:write. To avoid a
 # "pwn-request" exposure this job MUST:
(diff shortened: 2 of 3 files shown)


alxjrvs:
Added a changeset (`@randsum/games` patch → 3.0.1) to smoke-test OIDC publishing end-to-end.

**Test sequence after this merges:**
1. Merge this PR → main has the OIDC changes **and** a pending changeset.
2. Release workflow runs `changeset version` → opens a **"chore: version packages"** PR bumping games to 3.0.1 (no publish yet).
3. Merge that version PR → Release runs `bun run publish` via **OIDC** → games@3.0.1 publishes with provenance. ← the live OIDC test.

Verified locally with `changeset version` that the bump is isolated to games (roller/cli unaffected, no linked-version pull).

PR #1153, opened 2026-07-08
refactor(roller): unify duplicated modifier schemas into a single source

alxjrvs:
## What

Every one of the 19 roller modifiers had its notation schema implemented **twice**:

- `src/modifiers/<mod>.ts` — schema + `apply`/`validate` behavior + a `docs` array (drives `roll()`)
- `src/notation/definitions/<mod>.ts` — schema only (drives `parseModifiers`/`validateNotation`/`tokenize`)

The two copies had already drifted (e.g. `drop` parse used a `sumMatchCounts` helper in one and an inline `reduce` in the other; `compound`/`explode`/`penetrate` `toDescription` used different comparison formatters), and the main bundle measurably shipped **both**.

This PR makes `src/notation/definitions/<mod>.ts` the single canonical source of each modifier's `pattern`/`parse`/`toNotation`/`toDescription`. Each `src/modifiers/<mod>.ts` now **imports** its schema, spreads it, and attaches only dice‑pool behavior. Static docs moved to a dedicated pure‑data module (`src/docs/modifierDocData.ts`), so `dist/docs/index.js` no longer pulls in the roll engine and the `./tokenize` path stays schema‑only.

## Why

- **De‑duplication / drift** — one source of truth per modifier; the drift is reconciled deliberately (roll‑path semantics were treated as truth, with the full ~2700‑test suite incl. conformance vectors as referee).
- **Bundle correctness** — the main entry no longer ships two schema copies, and `./docs` is now genuinely "pure static data" (self‑contained, zero roll‑engine/parse identifiers), matching the package `CLAUDE.md` claim.

## Reconciled drift

- `compound`/`explode`/`penetrate` `toDescription` (object/condition form): the modifiers copy used `formatComparisonDescription` ("greater than or equal to 8"); the definitions copy used `formatComparisonNotation` (">=8"). Three tests (`conditional{Compound,Explode,Penetrate}`) pinned the richer description form, so the canonical schema adopts it. The public `modifiersToDescription` path already used the definitions copy; behavior is preserved elsewhere.
- `drop` parse (`sumMatchCounts` vs inline `reduce`) — semantically identical; kept the behavior‑free inline form on the canonical schema.

## Note on `sideEffects`

The task asked to set `"sideEffects": false`. This was attempted and **reverted**: it triggers a pre‑existing bunup/rolldown codegen bug that drops the re‑exported bindings from the barrel entry, producing an unloadable `dist/index.js` ("X is not declared in this file"). It reproduces on `origin/main` independent of this refactor (verified with `splitting` on and off, and with `minify` off). Left `true`; the consumer‑tree‑shaking win is deferred to a bunup upgrade. The de‑duplication already shrank the bundle substantially on its own.

## Verification

- **Tests**: full roller suite green — 2645 pass / 0 fail / 3 skip / 2 todo.
- **RDN conformance**: `@randsum/rdn conformance:check` produces a **byte‑identical** `v0.9.0.json` (no diff) — public notation behavior unchanged.
- **typecheck / lint / format / knip / arch‑check**: clean (all pre‑push hooks passed).
- **Bundle sizes** (brotli), budgets tightened to catch regressions:

  | File | Before | After | New budget |
  | --- | --- | --- | --- |
  | `dist/index.js` | 15.72 KB | **11.46 KB** | 16 → 13.25 KB |
  | `dist/docs/index.js` | 10.07 KB | **4.53 KB** | 20 → 5.25 KB |
  | `dist/tokenize.js` | 6.50 KB | 6.51 KB | 6.75 KB (unchanged) |
  | `dist/trace/index.js` | 909 B | 909 B | 5 → 1.1 KB |

  `dist/docs/index.js` is now self‑contained (no chunk imports) with zero roll‑engine identifiers.

- **Changeset**: `patch` for `@randsum/roller` (internal refactor; public surface unchanged, guarded by the public‑surface test).

From the 2026-07-07 monorepo deep-dive audit (Wave 2).


alxjrvs:
**Review: approved.** Reviewer verified all 19 schema unifications are semantically faithful (drop's two implementations were identical; the three reconciled toDescription drifts adopt the roll-path form), public surface unchanged, RDN conformance byte-identical, and dist/docs is now genuinely engine-free (index 15.7→11.5KB, docs 10.1→4.5KB brotli, budgets tightened). The benchmark-gate alert was investigated with an interleaved same-process A/B harness (21 trials, p50): parity within ±2% on all four benchmarks, hot path byte-identical to main — single-run-vs-cached-baseline runner drift, and the rerun passed. `sideEffects: false` deferred due to a bunup/rolldown bug reproducible on main (documented in PR). Follow-ups queued: bench-gate hardening (median-of-N), sumMatchCounts dead-code cleanup, docs-key sync test.

PR #1157, opened 2026-07-08
chore: consolidate root scripts and CI package jobs; harden the bench gate

alxjrvs:
## Summary

Wave 3 consolidation of root scripts and CI, plus a guard against the silent-skip that let dice-ui escape linting for months.

### Root scripts (`package.json`)
- **Deleted** the `check:all` alias (`check` is the real chain), the seven `discord:*` wrappers, and `dev` — `bun run --filter …` is the documented path for those.
- **THE GUARD — `scripts/check-workspace-scripts.ts`**: asserts every workspace package declares the standard set (`test`, `lint`, `format`, `format:check`, `typecheck`, `check`; plus `build` unless the package is source-only). `bun run --filter '*' <script>` **silently skips** packages that lack the target script — exactly how `dice-ui` escaped linting for months. The guard runs at the head of the root `check` chain **and** as its own CI job feeding the CI Gate, so a new package cannot silently opt out (`--no-verify` can't bypass the CI copy).
- Added a root `size` script (`--filter '*' size`). `size` stays **per-package optional / skip-tolerant** (only publishable packages define size-limit budgets); this is documented in the guard, which deliberately does **not** require `size`.
- `test` collapses to `--filter '*' test` (verified reliable, now also covers `site`).

#### Why `build`/`check` keep dependency ordering (not `--filter '*'`)
The audit suggested `--filter '*'` "runs in dependency order and skips missing scripts." Verified against bun 1.3.14 — it does **not**:
- `bun run --filter '*' build` **cold** exits 1: `cli`/`mcp`/`discord-bot` fail with `Could not resolve "@randsum/roller"` — `--filter` runs in parallel with no topological ordering.
- Parallel `bunup` **dts generation races** even when the dependency's `dist` already exists (reproduced: 6-way parallel tail fails on `mcp` while `mcp` builds fine alone).

So `build` and `check` retain dependency-ordered execution (roller → games → …). The guard makes the explicit per-package presence safe regardless of how the aggregate iterates.

### CI (`.github/workflows/ci.yml`)
- Folded the eight near-identical package jobs (roller, games, cli, mcp, discord-bot, site, rdn, dice-ui) into **one dynamic matrix**. `matrix` isn't available in a job-level `if` (actionlint confirms), so the `changes` job emits a matrix of only the **changed** packages via `jq`; each `include` leg carries its prereq builds, node version, coverage flag, and size gate. Prereqs build **sequentially** in the leg (games needs roller's dist first).
- **CI Gate** now needs the single `package` job (its `result` aggregates every leg — any leg `failure` ⇒ aggregate `failure`, so one entry still covers all eight) plus the new `workspace-scripts` job. `fail-fast: false` so each leg's real result reaches the gate.
- Scoped the **site** job's build to its dependency subtree (roller, games) instead of the full root build, matching siblings. (dice-ui is in the subtree but source-only, so nothing to build.)

### Bench gate
- `bench:ci` now runs the suite **3×** via `scripts/bench-ci.ts` and emits the **per-benchmark median** (kept as a wrapper so `roll.bench.ts` stays clean). Median rejects a single anomalous run without smearing a real regression the way a mean would.
- `alert-threshold` raised `150% → 160%` as belt-and-braces; `fail-on-alert` stays on.

### typecheck (item 4): left unchanged — deliberately
Switching root `typecheck` to `tsc -b` was evaluated and **rejected** because it does not demonstrably work as an equivalent:
- Composite build tsconfigs (`tsconfig.packages.json`) `include: src/**/*` and **exclude** `**/*.test.ts`; a `tsc -b` would stop typechecking test/benchmark files that the per-package `tsc --noEmit` currently covers (e.g. roller's config includes `__tests__`/`__benchmarks__`) — a coverage regression.
- Composite `tsc -b` **emits** `.d.ts` + `.tsbuildinfo` (unlike the current `--noEmit`) — a behavior change writing into `dist` during a typecheck.
- The astro apps (`rdn`, `site`) use `astro check` and extend `astro/tsconfigs/strict` (non-composite, rely on generated `.astro` types) — not `tsc -b`-buildable.

Per the task's "only if it demonstrably works; otherwise leave it alone and say so," `typecheck` is unchanged.

## Verification
- `bun run check` from a clean tree (dists wiped): **exit 0**, all packages green.
- The guard **passes**; temporarily stripping `lint` from `packages/mcp` made it fail with `@randsum/mcp: missing lint` and exit 1, then restored to green (demonstrated locally).
- `actionlint` passes on the edited workflow; `prettier --check` clean on all touched files.
- Pre-commit and pre-push hooks (build, codegen/conformance check, tests, audit, SCA, knip, arch) passed on push.

### Out of scope (parallel PRs own these)
Docs still referencing `check:all` (README.md, CONTRIBUTING.md, CLAUDE.md, PR template, skills, ADRs) are outside this PR's file scope; they should be updated to `check`.

From the 2026-07-07 monorepo deep-dive audit (Wave 3).


alxjrvs:
**Review: approved.** Matrix fold verified to preserve per-package gating (8 legs with original check-run names, fail-fast off, gate needs intact — required-check names unchanged); the new workspace-scripts guard demonstrably fails when a script is stripped; bench gate now median-of-3 with a 160% threshold, addressing the false positive that hit #1153.

PR #1160, opened 2026-07-08
refactor(roller)!: single-pass lexer for notation — validate/parse/tokenize/roll unified

alxjrvs:
## Summary

Replaces the five overlapping notation regex systems in `@randsum/roller` — the
strip-and-check validator, the per-schema merge scanner, the multi-pool splitter,
the hand-rolled `tokenize` pattern table, and the private die-type regexes in
`roll()` — with **one cursor-based lexer**. Net −721 / +962 lines (the old regex
systems are deleted; the new lexer + threaded semantics are added).

## Lexer / parser architecture (`src/notation/lexer/`)

- **`specs.ts`** — the single token inventory. `POOL_SPECS` (die-type heads:
  standard, percentile, fate, custom-faces, draw, geometric, zero-bias) and
  `MODIFIER_SPECS` (category + a sticky matcher per modifier). The modifier
  matchers are compiled **from the Wave-2 `notation/definitions/<mod>Schema`
  patterns** in RDN specificity order — the schemas remain the semantic source of
  truth for parse/format; the table only adds a stable token key, a faceted
  category, and disambiguation ordering.
- **`scan.ts`** — `scan(input)` is a pure left-to-right cursor pass emitting
  positioned `LexToken`s (`pool | modifier | annotation | repeat | unknown`). It
  enforces no structure; a pool is only recognized at a boundary (string start or
  after a `+`/`-` sign). This raw stream is exactly what `tokenize` exposes.
- **`parse.ts`** — `parseNotation(input)` layers the grammar over the token stream
  (must begin with a pool, no unknown tokens, positive-integer magnitudes, ≤1
  Count-family modifier, ≤1000 chars) and returns a typed pool AST plus the real
  error position. This is the **single acceptance authority**.

Every public surface is now a thin view over that one pass:

| Surface | Now |
| --- | --- |
| `isDiceNotation` | `parseNotation(...).valid` |
| `notation()` | throws `NotationParseError` with real `ErrorContext.position` |
| `validateNotation` | parse + describe; `error.position` populated |
| `notationToOptions` | AST → `ParsedNotationOptions[]` (multi-pool from the AST) |
| `tokenize` | positioned `scan` stream with descriptions attached |
| `roll()` `parseArguments` | consumes the same parser; special dice built from `dieType` |

## Deliberate behavior changes (before → after)

1. **Special-die semantics survive `notationToOptions`.** `ParsedNotationOptions`
   gains additive `dieType` / `fateVariant` / `customFaces`.
   - `notationToOptions('4dF')`: before → plain 3-sided pool; after →
     `dieType: 'fate'`, `fateVariant: 1`.
   - `notationToOptions('1d20+g6')`: before → dropped the special pool; after →
     both pools, second `dieType: 'geometric'`.
2. **`validateNotation` descriptions correct for special dice** (matching `roll()`):
   - `4dF`: `"Roll 4 3-sided dice"` → `"Roll 4dF"`
   - `dF.2`: `"Roll 1 5-sided die"` → `"Roll 1dF.2"`
   - `3DD6`: `"Roll 3 6-sided dice"` → `"Draw 3 from d6"`
   - `g6`: `"Roll 1 6-sided die"` → `"Roll g6 (geometric: roll d6 until 1)"`
   - `z6`: `"Roll 1 6-sided die"` → `"Roll 1z6"`
   - `d{fire,ice}`: `"Roll 1 2-sided die"` → `"Roll 1d{fire,ice}"`
   - `d%` and standard pools unchanged (percentile is genuinely d100).
3. **Error positions populated.** Thrown `NotationParseError`s and
   `validateNotation(...).error` now carry the real character offset
   (`ErrorContext.position` / `ValidationErrorInfo.position`); before → absent.
4. **Oversized input (>1000 chars) throws consistently.** Before →
   `notationToOptions` returned `[]` / silently split a long string while `roll()`
   threw; after → every public surface throws.
5. **Structural rejections.** A die/modifier out of pool position, an annotation
   before a pool (`[fire]1d20` → now invalid; `1d20[fire]` is still valid), or a
   leading-zero magnitude on a special die are now rejected. Bare `dN` is valid
   **only** at a pool boundary (start or after `+`/`-`); `4d6d6` and `2d62d6`
   throw; `4DD6L` throws at roll time (special dice reject modifiers). Acceptance
   of all well-formed notation is unchanged.
6. **Repeat operator (`xN`) scope is explicit.** `xN` repeats everything to its
   left, so it is accepted **only** as a trailing run, and a whole-string-trailing
   `xN` now repeats the **entire** expression.
   - `2d6+1d8x2`: before → the pre-lexer engine scoped a trailing `xN` to the last
     pool ("2d6 then two 1d8"); after → the whole expression twice ("(2d6+1d8) ×2",
     4 pools).
   - `4d6x2+2d8`: before → accepted but the `x2` was **silently dropped** (no
     effect); after → a positioned rejection (`isDiceNotation` → false, `roll()`
     throws `"…must be at the end of the notation"`).
   - `4d6Lx6` (single-pool trailing) and `2d6x2x3` (nested trailing) are unchanged.

Additionally (public-API hardening, not a behavior change for valid input): the
lexer is now an **iterative** cursor loop rather than per-token/per-char
recursion, so the `tokenize` subpath — which is not shielded by the 1000-char
parse gate — no longer overflows the stack on very long input (e.g.
`tokenize('q'.repeat(1_000_000))`). Error positions are also reported in the
caller's coordinates (the internal trim offset is added back), so `' 1d20&&'`
reports the `&` at its original offset.

Two `buildNotationPattern` unit tests were removed (the function is deleted); no
test that pins well-formed acceptance or special-die validation was loosened.

## Benchmark A/B (local, interleaved, p50 over 15 trials × 20k `roll()` calls/trial)

`origin/main` dist vs this branch's dist, alternating order each trial:

| notation | old p50 (ms) | new p50 (ms) | delta |
| --- | --- | --- | --- |
| `1d20` | 217.9 | 49.8 | **−77%** |
| `4d6L` | 361.6 | 197.5 | **−45%** |
| `2d20H` | 349.2 | 187.2 | **−46%** |
| `4d6R{1}` | 374.3 | 182.5 | **−51%** |

Two runs, stable to <1%. Dropping the per-roll giant-alternation validator is the
win. The lexer is linear → inherently ReDoS-free.

## Bundle size (brotli, min) — budgets shrink

| bundle | limit (was → now) | actual |
| --- | --- | --- |
| `dist/index.js` | 13.25 → **12.25 KB** | 11.82 KB |
| `dist/tokenize.js` | 6.75 → **6.5 KB** | 6.24 KB |
| `dist/docs/index.js` | 5.25 KB | 4.53 KB |
| `dist/trace/index.js` | 1.1 KB | 962 B |

## Verification

`typecheck`, full roller suite (2650 pass / 0 fail), root `test` (all consumer
packages green), `lint`, `knip`, `size`, and `@randsum/rdn conformance:check`
(**byte-identical** `v0.9.0.json`) all pass. `arch:check` is unaffected by this
change (it fails locally only on an unsupported node version; CI runs it on a
supported runtime).

Changeset: **major** for `@randsum/roller`.

## Note

This PR was completed across **two agent sessions** — a first session authored the
lexer, parser, and threaded most surfaces; this session assessed the inherited
state, finished the `validateNotation`/`optionsToDescription` special-die
descriptions, and ran the full verification + benchmark + size matrix.

From the 2026-07-07 monorepo deep-dive audit (finale).


alxjrvs:
**Review: approved after fixes.** Adversarial review ran a 610-case differential fuzz vs main (every disagreement maps to an enumerated deliberate change — and the branch *fixes* three internal-error crashes main had), 13/13 special-die round-trips via seeded RNG, and confirmed the −43–77% p50 perf claims. Two confirmed blockers were fixed in 393373fb: the recursive scanner stack-overflowed the public tokenize subpath at ~20k chars (now iterative, 1M-char regression tests added) and non-trailing `xN` repeat was silently dropped (now a positioned rejection; whole-expression trailing scope is a deliberate, documented change from main's last-pool scoping). Error positions now in caller coordinates. Suite 2666/0, conformance byte-identical, sizes within tightened budgets.

PR #1208, opened 2026-08-15
chore(ci): rename the aggregate gate display name to "CI Success"

### LINKED ISSUE #1205: Standardize aggregate CI gate job name to "CI Success"
Cross-repo standardization. Purely a rename — no change to what CI actually checks.

## The problem

Four active monorepos each landed on the same CI topology independently — a `changes` paths-filter job, parallel per-concern jobs, and one `if: always()` aggregate job that `needs:` every other job and is the sole required status check. All four named that final job differently:

| Repo | Job key | Display name |
|---|---|---|
| `randsum` | `ci-gate` | `CI Gate` |
| `binfinite-app` | — | `CI Success` |
| `SU-SRD` | `quality-checks` | *(none)* |
| `optfall` | `gate` | `gate` |

Because the required-check name is what branch protection matches on, four names means ruleset setup is bespoke per repo and the `agent-friendly-repo` skill cannot apply one checklist across them.

## The change

Rename this repo's aggregate job so its **display name is `CI Success`**, matching `binfinite-app` — chosen as the standard because it is already wired into that repo's deploy gating and branch protection, so it is the one with the most to lose from moving.

In `.github/workflows/ci.yml`, change the `ci-gate` job's `name:` from `CI Gate` to `CI Success`. Keep the job key.

## Sequencing matters

Update the required status check in the branch ruleset for `main` **in the same window** as the rename, or PRs will block waiting for a check that can no longer arrive:

1. Change the `name:` and merge to `main` so a run with the new name exists
2. Update the ruleset's required check to `CI Success`
3. Remove the old name from the ruleset

Note this repo has a Dependabot auto-merge workflow. Auto-merge waits on the required check, so it will stall on any PR opened during the window between steps 1 and 2 — keep that window short, or pause the workflow while it is open.

## Acceptance

- [ ] Aggregate job displays as `CI Success`
- [ ] Branch ruleset for `main` requires `CI Success`
- [ ] Dependabot auto-merge still lands a minor/patch PR end to end
- [ ] A PR blocks correctly when a downstream job fails

alxjrvs:
Closes #1205

Renames this repo's aggregate CI gate job so its **display name is `CI Success`**, matching `binfinite-app`. Cross-repo standardization across the four monorepos that share this CI topology (`randsum`, `binfinite-app`, `SU-SRD`, `optfall`), so `agent-friendly-repo` can apply one ruleset checklist instead of a bespoke one per repo.

## MERGE WARNING — the required status check must be updated by hand, in this order

The required-check name is what branch protection matches on. **This PR deliberately does not touch the ruleset**, so merging it renames the check that CI produces while the ruleset still requires the old name. Every PR then blocks forever waiting for a check that can no longer arrive.

Do this, in order:

1. **Merge this PR** so a run producing `CI Success` exists on `main`.
2. **Update the `main` ruleset's required status check to `CI Success`.**
3. **Remove `CI Gate` from the ruleset.**

Steps 2 and 3 are manual, human-only actions. Nothing in this PR performs them.

### Repo-specific hazard: Dependabot auto-merge will stall in the window

This repo has a Dependabot auto-merge workflow (`.github/workflows/auto-merge.yml`) that calls `gh pr merge --auto`. **Auto-merge waits on the required check.** Any Dependabot PR opened between steps 1 and 2 will sit with auto-merge armed against a check name that no longer arrives, and will silently never land.

Mitigation — pick one:

- **Keep the window short.** Do step 2 immediately after this merges, not "later today."
- **Or pause the workflow while the window is open** (`gh workflow disable auto-merge.yml -R RANDSUM/randsum`, re-enable after step 3). Safer if there is any gap between the merge and the ruleset edit.

Any PR already stalled when step 2 completes needs a nudge to re-evaluate — re-run it or push an empty commit.

## What changed

Purely a rename. **Nothing about what CI checks changes:** the job key stays `ci-gate`, the `needs:` list is untouched, and no job was added, removed, or reordered.

| File | Change |
|---|---|
| `.github/workflows/ci.yml` | `ci-gate` job's `name:` → `CI Success` (the only functional change); 3 prose references |
| `.github/workflows/workflow-lint.yml` | prose reference |
| `lefthook.yml` | prose reference |
| `osv-scanner.toml` | prose reference in the two-gates note |
| `.dependency-cruiser.cjs` | prose reference |
| `scripts/check-workspace-scripts.ts` | prose reference in the wiring docblock |

The prose references were updated so a grep for the gate still finds every place that describes it — a stale "CI Gate" in a comment is exactly the kind of thing that makes the next person doubt which name the ruleset wants. They read "the CI Success gate" because "the CI Success" alone is not a noun phrase; the word "gate" stays accurate since the job key is still `ci-gate`. A lowercase generic "the CI gate" in `scripts/sca-scan.sh` was left alone — it describes the concept, not the job's display name.

`actionlint` passes on both touched workflows.

## Note on CI: the `audit` job is red for an unrelated, pre-existing reason

Two **new** high advisories landed upstream since `main` last ran green (2026-08-11) and are not in the `--ignore` list: `nanoid` (GHSA-2v37-7h3g-55p8) and `extract-zip` (GHSA-jmr9-qjv8-65gv). `bun audit --audit-level=high` fails on `main` itself right now, so it fails here too and takes `CI Success` red with it.

**This is not caused by this PR** — the diff touches no dependency files (no `bun.lock`, no `package.json`). It needs its own fix, in the shape of #1197, and it blocks step 1 above: this PR cannot merge until the audit gate is green again.

### DIFF
--- a/.dependency-cruiser.cjs
+++ b/.dependency-cruiser.cjs
@@ -5,7 +5,7 @@
 //   2. no game subpath imports another game subpath
 //   3. no consumer reaches into @randsum/*/src or @randsum/*/dist
 //
-// Run:  bun run arch:check   (wired into lefthook pre-push and the CI Gate)
+// Run:  bun run arch:check   (wired into lefthook pre-push and the CI Success gate)
 
 /** @type {import('dependency-cruiser').IConfiguration} */
 module.exports = {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,8 +117,8 @@ jobs:
     if: fromJSON(needs.changes.outputs.matrix).include[0] != null
     runs-on: ubuntu-latest
     strategy:
-      # Run every package leg even if a sibling fails, so the CI Gate sees each
-      # package's real result instead of a fail-fast cancellation.
+      # Run every package leg even if a sibling fails, so the CI Success gate
+      # sees each package's real result instead of a fail-fast cancellation.
       fail-fast: false
       matrix: ${{ fromJSON(needs.changes.outputs.matrix) }}
     steps:
@@ -170,7 +170,7 @@ jobs:
         run: bun run --filter "@randsum/$PKG" size
 
   ci-aggregator:
-    # Guard against a job silently opting out of the CI Gate. `ci-gate`'s `needs:`
+    # Guard against a job silently opting out of the gate. `ci-gate`'s `needs:`
     # list is hand-maintained, and a job missing from it still runs, still goes
     # red, and still cannot fail the only required status check — the gate stops
     # gating with no symptom. This job diffs the two lists mechanically, and is
@@ -186,7 +186,8 @@ jobs:
     # Guard against a package silently opting out of the standard script set.
     # `bun run --filter '*' <script>` skips packages that lack the target script,
     # which is how dice-ui escaped linting for months (audit Wave 3). This job
-    # feeds the CI Gate so the escape hatch cannot be bypassed with --no-verify.
+    # feeds the CI Success gate so the escape hatch cannot be bypassed with
+    # --no-verify.
     name: workspace-scripts
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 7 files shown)


alxjrvs:
⚠️ **Correction: the merge sequence in this PR's body is backwards, and following it will not work.**

The body says to merge first, then update the ruleset. That is impossible — **this PR blocks itself.**

## Why

This PR *is* the workflow change, so its own CI run already reports the **new** check name. The ruleset still requires the **old** one, which therefore never arrives on this PR. Verified:

- This PR reports: `CI Success`
- The ruleset requires: the old name
- Result: `mergeStateStatus = BLOCKED`, with every check green

Merging cannot come first, because the merge is what is blocked.

## The correct order

1. **Update the ruleset's required status check to `CI Success` first.**
2. This PR immediately becomes mergeable — its run is already reporting that context.
3. Merge it.
4. **Rebase the other open PRs in this repo.** Between steps 1 and 3 they still report the old name and will block; a rebase onto the merged default branch picks up the renamed job and clears them.

There is no ordering that avoids a window entirely — requiring both names blocks in both directions, since whichever name a given run is not reporting never arrives. Step 4 is the cost, and it is a rebase rather than anything destructive.

## Note on the ruleset write

The ruleset change is a branch-protection write and needs to be made by a human — an agent's auto-mode classifier denies it, which is the correct outcome for this class of action. The exact command is in the session hand-off.

Everything else in this PR is unaffected; only the sequencing note above is wrong.

PR #1214, opened 2026-08-18
fix(discord-bot): make Sentry actually report, and stop syncCommands crashing the boot

alxjrvs:
## Why

Triaging "randsum is down" produced no signal anywhere it should have. Every other surface is healthy:

| Surface | State |
| --- | --- |
| randsum.dev / notation.randsum.dev | 200, all sitemap routes resolve |
| `@randsum/roller` / `games` / `cli` | resolve on the registry at 4.0.0 |
| CI on `main` | green |
| Render deploy `85e8fb32` (Aug 16) | reports **`success`** to GitHub |

A successful deploy plus a dead bot means a **runtime** failure, not a build failure. The place that should have explained it — Sentry — was empty, and that turns out to carry no information at all.

## The Sentry stub

`forwardToSentry` was an empty function body. `SENTRY_DSN` was read, stored, and logged as `enabled: true`, and then every capture went nowhere. The one env var whose entire job is explaining an outage has never sent an event.

This wires it to Sentry's envelope HTTP API using plain `fetch`. `@sentry/node` deliberately stays out of the dependency tree: the documented reason for avoiding it (bundle weight on a single-instance worker) still holds, and "POST one event" is all this bot needs. Three properties, each with a test:

- **Tracking never throws.** A delivery failure is warn-logged, never routed back through `captureException` (which would recurse). A malformed DSN degrades to logging-only and *says so at boot* — a typo'd DSN otherwise looks identical to a working one: the service comes up and silently reports nothing.
- **Captures survive a deliberate exit.** Delivery is async, so the fatal-login path — the single event that explains a crash loop — was being discarded by the immediate `process.exit(1)`. `index.ts` now awaits `flushErrorTracker()` first.
- **Delivery is injectable** via a `SendEnvelope` seam, mirroring the existing `RestLike` seam in `syncCommands`, so tests touch neither the network nor global `fetch`.

Verified end to end against a local ingest sink — with a bad token, the worker now delivers before exiting:

```
SINK_RECEIVED_URL /api/5/envelope/?sentry_key=testkey&sentry_version=7
SINK_EVENT {"exception":{"values":[{"type":"DiscordjsError [TokenInvalid]",
            "value":"An invalid token was provided."}]},"extra":{"phase":"login",...}}
```

## The boot-crash gap

`syncCommands` could take down the bot it explicitly promises never to take down. `commands.map(c => c.data.toJSON())`, `Routes.*` and `new REST()` all sat **above** the try block — and `toJSON()` validates the builder, so one over-long description throws. Its only call site is a top-level `await` in `index.ts`, so that rejection exits the worker and Render restarts it into a loop.

Moved inside the guard. **This is not the cause of the current outage** — all ten commands serialize fine today, which I checked — but it is the same shape as it, and the module's stated contract has to cover building the request, not just sending it.

## Triage runbook

Adds an outage-triage section to `apps/DEPLOY.md`. The most useful discovery: Render reports deploy state to the **GitHub deployments API**, so build health is readable with `gh` and no dashboard access. Knowing the build succeeded is what redirects triage from "what broke the deploy" to "what is failing at runtime" — the step that cost the most time here. Includes a table mapping where the boot log stops to what it means.

## What this does not do

It does not identify the root cause of the current outage — that needs the Render logs, which this session couldn't reach (the `render` MCP server is pending approval). Based on the code, a revoked/invalid `DISCORD_TOKEN` fits the evidence best: five backed-off retries then `exit(1)`, which Render turns into a restart loop. The runbook table maps that signature.

What it does do is make the next one self-reporting.

## Test plan

- `bun run --filter @randsum/discord-bot check` → exit 0 (build, typecheck, format, lint, test)
- 110 tests pass (8 new for the tracker, 1 new regression test for the serialize-crash path)
- `bun run knip` → clean

> Note: the pre-push `sca` step was excluded for this push — `osv-scanner` resolves to a mise shim with no version pinned, and Docker isn't running, so the scan can't execute locally. This branch changes no dependencies, and the CI `sca` job is the real gate.

### DIFF
--- a/apps/DEPLOY.md
+++ b/apps/DEPLOY.md
@@ -75,6 +75,15 @@ therefore sufficient — there is no separate registration step to remember.
 (Remove `DISCORD_GUILD_ID` to register globally — ~1h propagation; set it for instant
 per-guild registration during development.)
 
+`SENTRY_DSN` is a fourth `sync: false` env var and is **optional** — the tracker logs but does
+not report when it is unset, and the worker logs which mode it is in at boot
+(`errorTracker.init … enabled: true|false`). Set it in the Render dashboard to get exception
+reporting; check that boot line before concluding anything from an empty Sentry project.
+
+> Reporting was a no-op stub until #1211: `SENTRY_DSN` was read and logged as `enabled: true`,
+> but the forwarder had an empty body, so no event was ever sent. Any "nothing in Sentry"
+> observation from before that commit carries no information.
+
 > This used to be a manual `bun run deploy-commands`. It was forgotten after #1191 renamed
 > `/su` to `/salvageunion`, leaving Discord advertising a command the worker no longer had;
 > every invocation silently timed out for a week while the bot was healthy. The startup sync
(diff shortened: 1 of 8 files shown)


alxjrvs (commented):

---

Read the diff and the surrounding sources (`errorTracker.ts`, `syncCommands.ts`, `index.ts`, `loginWithBackoff.ts`, `render.yaml`, the sync test helpers). No prompt-injection or agent-addressed text in the diff — the comments are ordinary rationale comments.

The two core fixes are sound: moving `toJSON()`/`Routes.*`/`new REST()` inside the `try` genuinely closes the boot-crash path (the sole call site really is a top-level `await` at `apps/discord-bot/src/index.ts:102`), and `localCount: commands.length` is equivalent to the old `localPayload.length`. `parseDsn` produces the correct ingest URL for both hosted and path-prefixed DSNs, and query-string auth (`sentry_key`/`sentry_version=7`) is a valid Sentry ingest auth mode. `render.yaml` already carries `SENTRY_DSN` as `sync: false`, so the DEPLOY.md claim checks out.

Findings, none blocking:

- `apps/discord-bot/src/utils/errorTracker.ts:107` — `initErrorTracker` sets `trackerState.send` unconditionally but returns early on the no-DSN (`:111`) and invalid-DSN (`:117`) paths **without clearing `trackerState.target`/`trackerState.dsn`**, so a re-init that should disable reporting instead keeps POSTing to the previously-resolved ingest URL through the newly-installed transport. Harmless in the app (init runs once, `index.ts:17`), but it is why `__tests__/utils/errorTracker.test.ts:17` must cache-bust the module import: if Bun's resolver strips or rejects the `?case=` query on a `.js`-specified TS module, every case after the first shares module state and *"does not send when no DSN"* fails. Worth confirming the suite passes on the pinned Bun 1.3.14 rather than assuming; a `trackerState.target = undefined` on both early-return paths would make the isolation incidental instead of load-bearing.

- `apps/discord-bot/src/index.ts:65` — `shutdown()` calls `process.exit(0)` without `await flushErrorTracker()`. The module docblock at `errorTracker.ts:21-25` states "Captures survive a deliberate exit" as an invariant that "holds no matter what", and `CLAUDE.md` repeats it, but SIGTERM/SIGINT — the signal Render sends on every restart and deploy — is the second deliberate-exit site and drops any in-flight delivery (e.g. a `client.error` capture from `index.ts:53` that just fired). Either flush there too (the function would have to become async and be invoked via `void`) or narrow the documented invariant to the login path.

- `apps/discord-bot/src/utils/errorTracker.ts:169` — the stack is shipped as a raw string in `extra.stack` rather than as `exception.values[0].stacktrace.frames`, so Sentry issues arrive with no frames: no code context, no suspect-commit attribution, and grouping falls back to exception type + message. Functional, but it degrades the exact use case the PR is justified by ("the one event that explains a crash loop").

- `apps/discord-bot/src/utils/errorTracker.ts:128` — `trackerState.dsn` is assigned but never read anywhere after the switch to `target` (`captureException:206` and `forwardToSentry:191` both gate on `target`). Dead state; drop the field.

- `apps/DEPLOY.md:128` — the triage table row reads `login.retry ×5 then login.failed`, but `loginWithBackoff` emits `login.retry` only on attempts 1–4 and `login.failed` on attempt 5 (`loginWithBackoff.ts:19`, `:49-63`) — four retry lines, not five. An operator counting five lines in a runbook will conclude the log is truncated.

PR #1215, opened 2026-08-18
fix(discord-bot): make the gateway connection observable

alxjrvs:
## The incident

The bot was down. Every signal Render had said it was fine:

| Signal | Reported |
| --- | --- |
| Build / deploy | `succeeded` |
| `server_failed` events | **none**, across 7 days |
| Service status | live |
| Logs | identical `metrics.flush` heartbeats, every 5 min |

All of that was **also** consistent with a completely disconnected bot, because none of it was measuring the connection.

The starkest symptom, from the log tail:

```
02:18:29  bot.connecting
   … 60 minutes of metrics.flush, nothing else …
03:18:51  bot.login_succeeded
03:18:52  bot.ready   tag: RANDSUM.io#8392  guilds: 15
```

`client.login()` hung for **a full hour** and emitted not one line. There are zero `login.retry` warnings in that window — the hang was *inside* login, and `loginWithBackoff` only logs when a call **rejects**. An hour-long outage with no log line and no failed health check.

## Two gaps caused the invisibility

**1. No shard lifecycle handlers.** `index.ts` registered `ClientReady`, `InteractionCreate`, `GuildCreate` and `client.on('error')`. None of those fire on a shard disconnect — shard errors and disconnects are separate events in discord.js. So a dropped WebSocket was silent, the process stayed alive, and Render stayed green.

**2. The heartbeat was a false liveness signal.** `metrics.flush` is a bare `setInterval`. It proves the event loop is turning and nothing more. That's worse than no heartbeat: it *reads* as health while saying nothing about the connection. Hundreds of those lines are exactly what the outage looked like.

## The fix

- **`utils/gateway.ts`** wires `ShardReady` / `ShardResume` / `ShardReconnecting` / `ShardDisconnect` / `ShardError` to `gateway.*` log lines carrying shard id, close code, and how long the previous status held.
- **`metrics.flush` embeds `gatewaySnapshot()`.** The heartbeat now carries liveness:
  ```
  "gateway":{"status":"connecting","connected":false,"forMs":15652,"disconnects":0,"resumes":0}
  ```
  Grepping `'"connected":false'` finds an outage directly — no cross-line correlation.
- **A stall watchdog** reports a connection stuck off `ready` past 5 minutes, once per stall (a 5-minute outage should page you once, not 5 times). Started **before** login, because the failure it exists to catch is a login that never returns.
- **`bot.login_succeeded` records `elapsedMs`** — alone, that would have made the silent hour obvious.

**Deliberately no auto-restart on a stall.** Discord throttles session starts, and an hour-long login is consistent with exactly that — a watchdog that exited would deepen the hole it was climbing out of. It escalates to the error tracker and lets a human decide.

## Design note

The state machine is separated from the discord.js wiring, so it's testable without a gateway. The tests assert the **not-ready** paths specifically — a suite that only checked "ready works" would have passed against the broken code too, which is how this gap survived in the first place.

## Also

Corrects the triage runbook I added in #1214, which would have **misread this very incident**: it treated a green Render dashboard as evidence of health. Render only knows the process is alive. The runbook now leads with the heartbeat's `connected` field and carries an explicit warning against concluding "the bot is fine" from Render state alone.

## Test plan

- 118 tests pass (8 new)
- `bun run knip` → clean
- Booted the built worker under `node` and confirmed the new lines land: `gateway.connecting` on startup, and the heartbeat reporting `"connected":false` while disconnected

> Note: pre-push `sca` excluded — `osv-scanner` is a mise shim with no version pinned locally and Docker isn't running. No dependency changes in this branch; CI's `sca` job still gates.

### DIFF
--- a/apps/DEPLOY.md
+++ b/apps/DEPLOY.md
@@ -118,23 +118,43 @@ the first two need no Render access at all.
    `curl -sS https://registry.npmjs.org/@randsum/roller/latest` for the packages. "Randsum is
    down" is most often only one of these four.
 
-3. **Read the logs.** Render dashboard → `randsum-discord-bot` → **Logs**, or the `render` MCP
-   server (`.mcp.json`). The bot logs one JSON line per lifecycle event, so grep for the boot
-   sequence: `errorTracker.init` → `bot.connecting` → `login.retry` → `bot.login_succeeded` →
-   `bot.ready` → `commands.sync.*`. Where it stops tells you which failure this is:
+3. **Check the heartbeat first — it answers "is it connected?" directly.** Every
+   `metrics.flush` line (one per 5 minutes) embeds the gateway snapshot:
 
-   | Last line seen                     | Meaning                                                       |
-   | ---------------------------------- | ------------------------------------------------------------- |
+   ```
+   "gateway":{"status":"ready","connected":true,"forMs":812344,"disconnects":0,"resumes":0}
+   ```
+
+   Grep the logs for `'"connected":false'`. A hit is an outage, with `status` and `forMs`
+   saying which kind and for how long. **This is the fastest signal available and it should be
+   your first look** — it needs no correlation across lines.
+
+4. **Read the boot sequence.** Render dashboard → `randsum-discord-bot` → **Logs**, or the
+   `render` MCP server (`.mcp.json`). One JSON line per lifecycle event:
+   `errorTracker.init` → `bot.connecting` → `gateway.connecting` → `bot.login_succeeded`
+   (with `elapsedMs`) → `gateway.ready` → `commands.sync.*` → `bot.ready`. Where it stops tells
+   you which failure this is:
+
+   | Last line seen                      | Meaning                                                       |
+   | ----------------------------------- | ------------------------------------------------------------- |
    | `login.retry` ×5 then `login.failed`| Bad/revoked `DISCORD_TOKEN` → exit 1 → Render restart loop. Rotate the token (below). |
-   | `bot.ready`, then silence          | Connected and healthy; the problem is command registration or a specific command. |
-   | `commands.sync.failed`             | Registry desync only. The bot still serves its previous command list. |
-   | nothing at all                     | Service suspended, or the deploy never started. Check Render **Events**. |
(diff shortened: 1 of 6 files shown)


alxjrvs (commented):

---

Reviewed the diff against the surrounding code (`src/utils/gateway.ts`, `metrics.ts`, `index.ts`, `logger.ts`, `errorTracker.ts`, and the existing `errorTracker.test.ts` precedent for the cache-busted dynamic import). No text in the diff is addressed to a reviewing agent; the prose is developer-facing commentary.

The core mechanism is sound: the state machine is pure and testable, `registerGatewayLogging` matches discord.js v14's `ShardReady`/`ShardResume`/`ShardReconnecting`/`ShardDisconnect`/`ShardError` signatures, `metrics.ts` → `gateway.ts` introduces no import cycle, and both timers are `unref`'d.

Findings (none blocking):

- `apps/discord-bot/src/utils/gateway.ts:129` — the stall clock is `state.changedAt`, which `recordGatewayEvent` resets on *every* transition (line 108), so a flapping connection never accumulates 5 minutes in one status: discord.js emits `ShardDisconnect`/`ShardReconnecting` per reconnect attempt, and attempts land well under the threshold, so `gateway.stalled` and its `captureException` escalation never fire for a bot stuck in a reconnect loop — one of the two outage shapes this module exists to catch. Tracking a `notReadySince` timestamp (cleared only on `ready`) instead of time-in-current-status would close it; the log lines and `connected:false` heartbeat still expose the outage, which is why this is not blocking.
- `apps/discord-bot/src/utils/gateway.ts:127` — `alarmed` is cleared only on a transition to `ready` (line 109), so after one alarm any further degradation is silent until the connection actually recovers: an alarm raised while `connecting` suppresses escalation for a subsequent permanent `disconnected` state indefinitely. That is defensible as "one page per outage", but combined with the reset above it means the Sentry path is effectively one-shot per process.
- `apps/discord-bot/__tests__/utils/gateway.test.ts:74` — the watchdog tests assert only `not.toThrow()` and never that `gateway.stalled` was logged, so the module's headline behaviour is untested: a `checkGatewayHealth` that returned early in every case would pass all three tests, including the one named "a stall alarms once" (line 89), whose "second call must be a no-op" comment has no assertion behind it. The heartbeat test at line 107 shows the pattern that would work here — capture `console.error` and assert one `gateway.stalled` line for two calls.

PR #1217, opened 2026-08-18
feat(discord-bot): Discord-webhook alerting and a dead-man's-switch heartbeat

alxjrvs:
**Phase 0 of the Cloudflare migration — and the only phase that's urgent.** This doesn't depend on Cloudflare, doesn't depend on the migration happening at all, and works from Render today.

## Why

The bot has had **no alerting of any kind**. #1214 made Sentry reporting actually work and #1215 made the gateway observable, but both write somewhere nobody is watching — `SENTRY_DSN` has never been set in production, and log lines only help someone already reading logs.

That's why 2026-08-18 was an hour of a healthy-looking process with every dashboard green.

## Two sinks, because they catch different failures

**`DISCORD_ERROR_WEBHOOK_URL`** — posts captured exceptions to a Discord channel. No third-party account, and it reuses infrastructure already running: the bot *is* a Discord bot, so the delivery channel was always sitting there. Runs alongside `SENTRY_DSN` rather than instead of it, so changing sinks is a config change with an overlap rather than a cutover.

The **dedupe is load-bearing, not tidiness**. Sentry fingerprints server-side, so a crash loop there is one issue with a rising count. A webhook has no such thing, and Discord rate-limits a webhook to 5 requests per 2 seconds — so without suppression, the failure that *most* needs reporting (a tight restart-crash loop) is exactly the one that floods the channel and gets itself rate-limited into silence. Signature is type + message, deliberately excluding stack and context, since a loop emits the same error with a fresh interaction id each time.

The structured log line is always emitted first: **dedupe governs notification, never the record.**

**`HEALTHCHECK_URL`** — the piece neither Sentry nor Cloudflare can provide, and the one that would actually have caught this morning. Every other signal reports from *inside* the process: the error tracker needs something to throw, the watchdog needs the event loop turning. A process that is dead, killed, or never restarted emits nothing. And a gateway bot has no inbound endpoint to poll, so liveness must be **pushed, not pulled**.

Started before the client exists, so a death during boot is covered too, and pings immediately on start so a restart is visible without waiting an interval.

## Coverage is now complete

| Failure | Caught by |
| --- | --- |
| Something threw | webhook |
| Gateway stalled, process alive | #1215 watchdog manufactures an exception → webhook |
| Process dead entirely | pings stop → monitor alerts |

That third row has never been covered at any price.

## Verified end to end

Booted the built worker against a local sink with a bad token:

```
HEARTBEAT GET /hc/abc
WEBHOOK title= ⚠️ DiscordjsError [TokenInvalid] fields= [{"name":"phase","value":"login"}]
```

One webhook post, not five — the retries log, and only the final failure captures.

## Safety

Neither path can become an outage. A delivery failure is warn-logged, never routed back through `captureException` (which would recurse). A heartbeat failure is deliberately *not* captured either: if the network is down the error path is down too, and the heartbeat's **absence** is already the alert.

## Test plan

- 126 tests pass (10 new: dedupe, distinct-errors-not-suppressed, webhook delivery, webhook failure tolerance, and 4 heartbeat cases)
- Fixed an env leak in the existing tracker tests — one case left `DISCORD_ERROR_WEBHOOK_URL` set and a later test attempted a real network call

## After merging

Set `DISCORD_ERROR_WEBHOOK_URL` and `HEALTHCHECK_URL` in the Render dashboard (both `sync: false`, declared in `render.yaml`). Until then the bot logs exactly as before and neither sink fires.

### DIFF
--- a/apps/discord-bot/.env.example
+++ b/apps/discord-bot/.env.example
@@ -11,6 +11,19 @@ DISCORD_CLIENT_ID=your_client_id_here
 # Remove or leave empty for global commands (takes up to 1 hour to propagate)
 DISCORD_GUILD_ID=your_guild_id_here
 
+# Optional: Discord webhook that captured exceptions are posted to. This is the
+# lowest-friction alerting path — no third-party account, and it reuses a channel
+# you already have. Repeated identical errors are suppressed for 10 minutes, so a
+# crash loop cannot flood the channel or trip Discord's 5-requests-per-2-seconds
+# webhook rate limit. Can run alongside SENTRY_DSN; both sinks fire independently.
+DISCORD_ERROR_WEBHOOK_URL=
+
+# Optional: dead-man's-switch URL, pinged every 60s (e.g. Healthchecks.io).
+# This is the ONLY signal that survives the process dying — an error tracker
+# needs something to throw, and a dead process throws nothing. A gateway bot has
+# no inbound endpoint to poll, so liveness has to be pushed rather than pulled.
+HEALTHCHECK_URL=
+
 # Optional: Sentry DSN for error tracking. When set, captured exceptions are
 # structured-logged AND delivered to Sentry over its envelope HTTP API
 # (src/utils/errorTracker.ts) — no @sentry/node dependency, so the worker bundle
--- a/apps/discord-bot/__tests__/utils/errorTracker.test.ts
+++ b/apps/discord-bot/__tests__/utils/errorTracker.test.ts
@@ -1,8 +1,11 @@
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
-import type { SendEnvelope, SendOutcome } from '../../src/utils/errorTracker.js'
+import type { PostDiscord, SendEnvelope, SendOutcome } from '../../src/utils/errorTracker.js'
 
 interface TrackerModule {
-  readonly initErrorTracker: (options?: { readonly send?: SendEnvelope | undefined }) => void
+  readonly initErrorTracker: (options?: {
+    readonly send?: SendEnvelope | undefined
+    readonly postDiscord?: PostDiscord | undefined
+  }) => void
   readonly captureException: (error: unknown, context?: Record<string, unknown>) => void
   readonly flushErrorTracker: () => Promise<void>
 }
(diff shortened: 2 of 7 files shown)


alxjrvs (commented):

---

Reviewed the diff against the current sources (`errorTracker.ts`, `heartbeat.ts`, `index.ts`, `logger.ts`, command paths). No text in the diff attempts to address a reviewing agent. Findings, most severe first — none blocking:

- `apps/discord-bot/src/utils/errorTracker.ts:330` — the new 10-minute dedupe gates **both** sinks, not just Discord (the constant's own doc says it is "load-bearing for the Discord sink specifically"), so Sentry now receives one event per signature per 10 minutes instead of every occurrence. Concrete case: a flapping gateway raising the same `ShardError` message every ~20s currently produces N Sentry events with distinct `shardId`/timestamps; after this change Sentry sees one, its occurrence count and `lastSeen` stop tracking the incident, and an ongoing outage reads as a single stale event. The same applies to per-interaction captures, whose `userId`/`interactionId` context is discarded for every suppressed repeat. If the intent is webhook-only protection, move the `isDuplicate` check into `forwardToDiscord`.

- `apps/discord-bot/src/utils/heartbeat.ts:99` — `startHeartbeat()` runs at module load (before `client.login()`) and pings immediately, so a boot-time crash loop keeps the dead-man's switch green. With a bad `DISCORD_TOKEN`, `loginWithBackoff` exhausts in ~15s (1+2+4+8) and `index.ts` exits 1; Render restarts, the process pings again on the next boot, and a Healthchecks.io period of 60s never fires because pings keep arriving at the restart cadence. The failure mode the module's own docblock names ("never restarted after a failed deploy", the 2026-08-18 disconnect) is precisely the one this ordering hides. Pinging from `ClientReady`/after `loginWithBackoff` resolves — or gating the interval on `gatewaySnapshot().connected` — would make the signal mean "alive *and* usable".

- `apps/discord-bot/__tests__/utils/heartbeat.test.ts:194` — the test named "pings immediately on start" calls `hb.sendHeartbeat()` explicitly right after `startHeartbeat`, then asserts `pinged.length >= 1`, so deleting the `void sendHeartbeat()` from `startHeartbeat` leaves the test green. The one behaviour it claims to protect is untested; asserting `pinged` before the manual call (or asserting length 2 after) would actually pin it.

- `apps/discord-bot/src/utils/errorTracker.ts:287` — the stack is interpolated into a triple-backtick fence with no escaping, so any error text containing ``` ends the code block and the remainder renders as arbitrary markdown in the alert channel. Discord API error messages and third-party error strings are not fully under this bot's control, so a crafted message can make an alert embed display attacker-chosen formatted text. Stripping or escaping backticks before the fence closes it.

- `apps/discord-bot/src/utils/errorTracker.ts:302` — dedupe bounds only *identical* signatures; distinct errors each POST immediately with no queue or throttle, so a burst of 6+ different errors in 2 seconds (e.g. an `uncaughtException` storm with varying messages) still trips Discord's 5-requests-per-2-seconds webhook limit. Failures are logged as `errorTracker.discord_failed` and dropped — no `Retry-After` handling — so the alerts lost are exactly the ones from the busiest failure window.

- `apps/discord-bot/src/utils/heartbeat.ts:110` — `stopHeartbeat()` clears the timer but leaves `state.url` and `state.failures` set, so a post-shutdown `sendHeartbeat()` still pings and a later `startHeartbeat()` with `HEALTHCHECK_URL` unset silently keeps the old URL. Minor today (only `index.ts` shutdown calls it), but the "stop is idempotent" test does not cover it. Separately, `apps/discord-bot/CLAUDE.md`'s Environment Variables table is not updated with `DISCORD_ERROR_WEBHOOK_URL` or `HEALTHCHECK_URL`, though `.env.example` and `render.yaml` are.

PR #1218, opened 2026-08-18
feat(rdn): Cloudflare Workers deploy for notation.randsum.dev

alxjrvs:
**Phase 1 of the Cloudflare migration.** Nothing user-facing changes: DNS still points at Netlify, so this publishes *alongside* the live site rather than replacing it. The cutover is Phase 2.

`rdn` goes first deliberately — it's the lowest-stakes surface in the repo (pure static, no adapter, no server routes, one workspace dep), so its real job is proving the deploy path and CI wiring on something that can't break much.

## No Astro adapter, on purpose

`@astrojs/cloudflare` exists to run server-rendered routes, and this site has none. Adding it would introduce a Worker script, a `wrangler` peer dependency and an Astro version floor for zero behavioural gain.

`output: 'static'` + an assets binding is the entire deployment. Cloudflare serves `dist/` directly, and **requests to static assets are free, unbilled, and don't count against the Workers request quota**. `apps/site` is the case that genuinely needs an adapter; this isn't.

`not_found_handling: "404-page"` is set so unmatched routes get Astro's own 404 rather than Cloudflare's bare platform error — without it, every typo'd URL silently loses the site's styling and navigation.

## Why the deploy job lives here, not in Workers Builds

This repo has never had a deploy job, because Netlify deploys itself from its own git integration. Cloudflare's equivalent (Workers Builds) has a known flakiness report for workspace dependency resolution in monorepos and nothing authoritative for Bun — and `rdn` imports `@randsum/roller` through a workspace subpath export that must be built first or the Astro build resolves nothing.

CI already sequences that correctly, so deployment goes where the build is already known-good. Path-filtered per site, which buys back the same economy the `netlify.toml` `ignore` predicates were providing.

## Verified

Deployed by hand: **all 13 assets uploaded, `randsum-rdn` Worker created.**

## ⚠️ Two account steps before this can actually serve

Neither is code, and neither blocks merging:

1. **Register a `workers.dev` subdomain** — the account has none, so the Worker currently has no URL at all. One-time, interactive: https://dash.cloudflare.com/f5f08e7e86ab8c183e381d4504bf8ba5/workers/onboarding
2. **Add a `CLOUDFLARE_API_TOKEN` repo secret** — needs `Workers Scripts: Edit`. Until it exists, the workflow will run and fail on merge. Create it in the Cloudflare dashboard, then `gh secret set CLOUDFLARE_API_TOKEN`.

Until (1), there's no URL to smoke-test against before Phase 2 moves DNS — which is a real loss of safety, so it's worth doing before the cutover rather than after.

## Test plan

- YAML validated; `workflow-lint` (actionlint + zizmor) runs on this PR
- Action SHA-pinned to match the rest of the repo's workflows
- `concurrency` is `cancel-in-progress: false` — a cancelled half-finished asset upload is worse than a short queue

### DIFF
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -1,0 +1,48 @@
+# Deploy static sites to Cloudflare Workers on merge to main.
+#
+# Netlify deploys itself from its own git integration, which is why this repo
+# has never had a deploy job. Cloudflare has no equivalent — Workers Builds
+# exists, but its monorepo handling has a known flakiness report for workspace
+# dependency resolution and nothing authoritative for Bun, and this repo has
+# cross-package build ordering (roller -> rdn) that CI already sequences
+# correctly. So deployment lives here, where the build is already known-good.
+#
+# Path-filtered per site rather than deploying everything on every merge, which
+# is the same economy the netlify.toml `ignore` predicates were buying.
+name: Deploy (Cloudflare)
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'apps/rdn/**'
+      - 'packages/roller/**'
+      - 'bun.lock'
+      - '.github/workflows/deploy-cloudflare.yml'
+
+permissions:
+  contents: read
+
+concurrency:
+  # Never let two deploys race to the same Worker. Not cancel-in-progress: a
+  # half-finished asset upload that gets cancelled is worse than a short queue.
+  group: deploy-cloudflare
+  cancel-in-progress: false
+
+jobs:
+  rdn:
+    name: notation.randsum.dev
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: ./.github/actions/bun-setup
+
+      # rdn imports @randsum/roller via a workspace subpath export, so roller
+      # must be built first or the Astro build resolves nothing.
+      - name: Build
+        run: bun run --filter @randsum/roller --filter @randsum/rdn build
+
+      - name: Deploy
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+        run: bunx wrangler@4 deploy -c apps/rdn/wrangler.jsonc
--- a/apps/rdn/CLAUDE.md
+++ b/apps/rdn/CLAUDE.md
@@ -111,6 +111,13 @@ keeping the published JSON in sync with the source.
   `@astrojs/language-server`'s `getTsconfig`. Revisit when `@astrojs/check` ships a
   release whose peer range admits TypeScript 7.
 - Private app, never published to npm.
-- Deployed to Netlify (see `netlify.toml`).
+- **Deployed to both Netlify (`netlify.toml`) and Cloudflare Workers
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):

---

Read the diff and checked it against the repo's existing deploy config (`apps/rdn/netlify.toml`, `.github/workflows/ci.yml`, `render.yaml`, `apps/DEPLOY.md`).

No prompt-injection or instructions-to-the-reviewer text in the diff.

**Findings**

- `.github/workflows/deploy-cloudflare.yml:49` — The build uses a single combined `bun run --filter @randsum/roller --filter @randsum/rdn build`, while every other path in this repo that has the same roller→dependent ordering requirement sequences with `&&` (`apps/rdn/netlify.toml:3`, `apps/site/netlify.toml:7`, root `package.json:57`), and `.github/workflows/ci.yml:134-138` states outright that "bun's `--filter` runs in parallel without topological ordering … a combined multi-filter build would race." If that comment is accurate, the Astro build can start before `packages/roller/dist` exists and fail to resolve `@randsum/roller/docs` (used by `ModifierReference.astro`), producing an intermittently broken deploy — and the step's own comment ("roller must be built first") asserts a guarantee the command doesn't enforce. Counter-evidence: `render.yaml:15` ships the same combined form for discord-bot and reportedly deploys fine, so bun may in fact order topologically and the CI comment may be stale. Either way the two claims can't both be right; using the `&&` form here matches the `netlify.toml` this workflow is meant to mirror and removes the question. Not marked blocking because the render.yaml evidence suggests the race may not be real.
- `.github/workflows/deploy-cloudflare.yml:52-54` — No `CLOUDFLARE_ACCOUNT_ID` is exported. `wrangler deploy` runs non-interactively here, and if the API token can reach more than one Cloudflare account wrangler aborts with "more than one account available but unable to select one in non-interactive mode"; the deploy then fails on every run with nothing in the config to fix it.
- `.github/workflows/deploy-cloudflare.yml:54` — `bunx wrangler@4` resolves a floating version at deploy time and executes it in a step that has `CLOUDFLARE_API_TOKEN` in the environment. This repo pins every action by SHA and pins its other `bunx` tools exactly (`bunx publint@0.3.21`, `bunx @arethetypeswrong/cli@0.18.4`, `ci.yml:286-287`), so this is the one unpinned code-fetch-and-run with a production credential in scope.
- `.github/workflows/deploy-cloudflare.yml:23-27` — The path filter is narrower than the Netlify `ignore` predicate it claims parity with: `netlify.toml:11` also rebuilds on `package.json`, `tsconfig.json`, and `tsconfig.packages.json`, and neither list covers `.github/actions/bun-setup/**`. A root build-config change therefore republishes Netlify but not Cloudflare, so the two deploys silently diverge during exactly the migration window this PR creates.
- `apps/rdn/wrangler.jsonc:92` — `"$schema": "node_modules/wrangler/config-schema.json"` points at a file that never exists: wrangler is not a dependency anywhere in the workspace (only `@astrojs/cloudflare` is, in `apps/site`), and the deploy fetches it via `bunx`. Editor schema validation silently does nothing. Cosmetic only.
- `apps/DEPLOY.md:11-20` — The operations runbook is not updated. Its surface map still lists `apps/rdn` as Netlify-only, its config-sources note omits `wrangler.jsonc` and the new workflow, and there is no rollback procedure for the Cloudflare deploy. `apps/rdn/CLAUDE.md` got the dual-deploy note; the runbook that an on-call reader would actually open did not.

PR #1219, opened 2026-08-18
feat(site): Cloudflare Workers build target for randsum.dev

alxjrvs:
**Phase 3 of the Cloudflare migration.** Nothing user-facing changes — Netlify remains the default target and is still serving randsum.dev.

Astro allows exactly one adapter, so this can't be a swap: swapping outright would break the deploy currently serving the site, before the Cloudflare one has been proven on anything. `DEPLOY_TARGET` selects the adapter instead, defaulting to Netlify, so **both targets build from the same commit** and the cutover later is a one-line change plus deleting the loser.

## Four things the research got wrong

All found by building rather than reading docs.

**1. The Astro bump is not optional.** `@astrojs/cloudflare@14.2.0` declares `astro: ^7.0.0`, which admits our pinned 7.1.0 — but its code imports `beginContentEntryCollection` from `astro/app`, a symbol that only exists in **7.2.0+**. The declared peer range is simply wrong, and the build fails with a `Missing export` that reads like a bundler problem. Hence astro 7.2.2 + adapter 14.2.1.

**2. `main` must not be set in `wrangler.jsonc`.** The vite plugin validates it *during* the build and rejects a path that doesn't exist yet — which the build output can't, because the build is what produces it. Setting it deadlocks. The adapter supplies its own entrypoint and emits `dist/server/wrangler.json` fully resolved; **that generated file is what gets deployed**, and the hand-written one is build input only.

**3. The adapter prerenders in a workerd sandbox, which refuses WebAssembly.** Shiki's Oniguruma highlighter is Wasm-backed, so every page with a code block failed to render — most of a docs site — and the markdown pipeline separately broke on `require('path')`.

`prerenderEnvironment: 'node'` fixes both. Pointing Shiki at its pure-JS regex engine also works and was tried first, but it changes how code is highlighted *on a live site*; prerendering in Node is the same environment Netlify already uses, so both targets produce identical pages.

**4. `compatibility_date` must not be "today".** A date newer than the workerd bundled with the installed wrangler makes `wrangler dev` refuse to start — costing the ability to test locally, which is the only way to verify anything before DNS moves.

## Verified by running the real Worker in workerd

Not just building:

| Check | Result |
| --- | --- |
| Home, docs pages | 200 |
| Legacy 301s | resolve — **22 redirect rules parsed** from `astro.config`, no hand-authoring |
| `/schemas/*` | `application/json; charset=utf-8` via the new `_headers` (2 header rules parsed) |
| 404 page | returns 404 |
| `POST /api/roll` | correct roll — `4d6L` → 3 dice, drop-lowest applied |
| Bad notation | proper JSON error |

`bun run --filter @randsum/site check` also passes, so the Netlify path is intact.

## Notes

- **`_headers` duplicates the `netlify.toml` rule on purpose**, until Netlify is decommissioned. The format is understood by both hosts, so one file keeps the schema content-type correct on whichever is serving. Deleting the toml block would have been tidier and riskier — changing behaviour on the live site to fix one nobody is using yet.
- The session KV namespace the adapter configures is **auto-provisioned on deploy**, so it needs no manual setup despite the account having none today.
- `.wrangler/` added to `.gitignore` (local dev state).

## Still blocked on account setup

Same two as #1218, neither code:
1. **Register a `workers.dev` subdomain** — no URL exists otherwise
2. **Add `CLOUDFLARE_API_TOKEN`** repo secret (`Workers Scripts: Edit`) — until then the deploy job fails on merge

### DIFF
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -16,7 +16,8 @@ on:
     branches: [main]
     paths:
       - 'apps/rdn/**'
-      - 'packages/roller/**'
+      - 'apps/site/**'
+      - 'packages/**'
       - 'bun.lock'
       - '.github/workflows/deploy-cloudflare.yml'
 
@@ -46,3 +47,31 @@ jobs:
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
         run: bunx wrangler@4 deploy -c apps/rdn/wrangler.jsonc
+
+  site:
+    name: randsum.dev
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: ./.github/actions/bun-setup
+
+      # DEPLOY_TARGET selects the Astro adapter. Astro allows exactly one, and
+      # Netlify is still serving randsum.dev, so the config defaults to Netlify
+      # and only this job opts into Cloudflare. Both targets are buildable from
+      # the same commit until the cutover.
+      - name: Build
+        env:
+          DEPLOY_TARGET: cloudflare
+        run: |
+          bun run --filter '@randsum/roller' build
+          bun run --filter '@randsum/games' build
+          bun run site:build
+
+      # Deploy the adapter-GENERATED config, not apps/site/wrangler.jsonc. The
+      # hand-written one is build input (name, compat flags, assets policy); the
+      # adapter reads it and emits dist/server/wrangler.json with the resolved
+      # entrypoint and asset paths, which is the deployable artifact.
+      - name: Deploy
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+        run: bunx wrangler@4 deploy -c apps/site/dist/server/wrangler.json
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,6 @@ packages/site/public/api/
 
 # Codegen test harness temp files
 packages/games/.test-tmp/
+
+# Wrangler local state (dev sessions, cached bindings) — never committed.
+.wrangler/
--- a/apps/site/astro.config.mjs
+++ b/apps/site/astro.config.mjs
@@ -8,6 +8,7 @@ import starlight from '@astrojs/starlight'
 import starlightPageActions from 'starlight-page-actions'
 import starlightSidebarTopics from 'starlight-sidebar-topics'
 import netlify from '@astrojs/netlify'
+import cloudflare from '@astrojs/cloudflare'
 import react from '@astrojs/react'
 import { copyMarkdownToDist } from './src/integrations/copy-markdown-to-dist'
 import { copySchemaToDist } from './src/integrations/copy-schema-to-dist'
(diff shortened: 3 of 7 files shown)


alxjrvs (commented):

---

I read the full diff and checked it against the repo (workflow, `astro.config.mjs`, the site's Astro integrations, `netlify.toml`, `ci.yml`, the adapter's own type definitions, and the local `.wrangler/deploy/config.json` which confirms `dist/server/wrangler.json` is the generated artifact path).

No prompt-injection or instructions-to-the-reviewer text in the diff; the comments are ordinary rationale prose.

**Findings**

- `apps/site/astro.config.mjs:33` — Selecting the Cloudflare adapter silently disables the `copy-markdown-to-dist` integration: it looks for `<dir>/../.netlify/build`, which under the Netlify adapter is `apps/site/.netlify/build` but under Cloudflare resolves to `apps/site/dist/.netlify/build` (`dir` becomes `dist/client`) and does not exist, so the hook returns early. The raw `.md` files that `starlight-page-actions` links to are produced by `vite-plugin-static-copy` into the server build dir, so they will not be under the served assets on the Cloudflare deploy — every "view as Markdown" link 404s there. Harmless while Netlify serves `randsum.dev`, but it breaks at cutover unless the integration is made target-aware. Not blocking.

- `apps/site/wrangler.jsonc:192` — `assets.directory: "./dist"` points at the build root, which contains `dist/server/` (worker bundle and sourcemaps) alongside `dist/client/`. The deploy step correctly uses the adapter-generated config, so production is unaffected — but the sibling `rdn` job three lines above in the same workflow deploys with `-c apps/*/wrangler.jsonc`, and any copy of that shape here (or a hand-run `wrangler deploy -c apps/site/wrangler.jsonc`) would publish the server bundle as public static assets. Narrow it to `./dist/client` or drop the key, since the comment says the adapter resolves it anyway. Not blocking.

- `.github/workflows/deploy-cloudflare.yml:20` — `paths:` is workflow-scoped, not job-scoped, so every listed path runs *both* jobs: broadening `packages/roller/**` to `packages/**` plus adding `apps/site/**` means a site-only change now redeploys `notation.randsum.dev`, and an `apps/rdn/**` change redeploys `randsum.dev`. The header comment's claim of "path-filtered per site" does not hold; per-job filtering needs a `dorny/paths-filter` job like `ci.yml` already uses. Not blocking (wasted deploys, not incorrect output).

- `.github/workflows/deploy-cloudflare.yml:62` — Nothing exercises `DEPLOY_TARGET=cloudflare` before merge: `ci.yml`'s site leg builds with the default (Netlify) adapter, so a Cloudflare-only build break (the Wasm/`require('path')` class of failure the comments describe) first appears as a failed deploy on `main`. A PR-time build with the flag set would close that gap. Not blocking.

- `apps/site/CLAUDE.md:6` — Still states the site "deploys to Netlify on push to `main`" with no mention of the Cloudflare target, while `apps/rdn/CLAUDE.md` documents its equivalent dual-deploy in detail. Stale as of this change. Not blocking.

The astro `7.1.0 → 7.2.2` bump is safe against the declared peers (`@astrojs/netlify@8.1.2` wants `astro ^7.0.0`, `astro-expressive-code@0.44.0` admits `^7.0.0`, `@astrojs/cloudflare@14.2.1` wants `^7.2.0`), `prerenderEnvironment: 'node'` is a real option in `@astrojs/cloudflare@14.2.1`, and the lockfile's `path-to-regexp@6.3.0` / `@jridgewell/trace-mapping@0.3.9` hoists are patched versions with the newer ones retained as nested entries — no advisory regression.

PR #1220, opened 2026-08-18
fix(ci): skip Cloudflare deploy when the API token is absent

alxjrvs:
The deploy workflow added in #1218 **failed on merge to main** — `CLOUDFLARE_API_TOKEN` doesn't exist yet — and would keep failing on every merge until it does.

A check that is red for a known, documented, expected reason is worse than one that does nothing. It trains people to ignore a failing check, and the next *real* failure hides inside the noise. This repo's own conventions make that argument elsewhere: an advisory check that can never meaningfully pass "teaches people to skim the whole checks list."

Both jobs now skip with a `::notice::` instead, and begin deploying the moment the secret is added — no further change needed here.

```
::notice::CLOUDFLARE_API_TOKEN is not set — skipping deploy.
Add the secret to enable Cloudflare deploys. Netlify is unaffected.
```

Netlify continues to serve both sites regardless; nothing about this affects the live deploys.

### DIFF
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -9,6 +9,13 @@
 #
 # Path-filtered per site rather than deploying everything on every merge, which
 # is the same economy the netlify.toml `ignore` predicates were buying.
+#
+# Both jobs SKIP rather than fail when CLOUDFLARE_API_TOKEN is absent. That is
+# the state of the world until the secret is added, and a workflow that fails
+# red on every merge for a known, documented, expected reason is worse than one
+# that does nothing: it trains people to ignore a failing check, and the next
+# real failure hides inside the noise. Adding the secret enables deploys with no
+# further change here.
 name: Deploy (Cloudflare)
 
 on:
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

**Review — `fix(ci): skip Cloudflare deploy when the API token is absent`**

The mechanism is right: `secrets` is not an available context in either job-level or step-level `if:`, so a shell guard inside `run:` is the only way to branch on a secret's presence. The check is correctly quoted, and the default runner shell (`bash -e`) does not set `-u`, so the unset-variable read is safe.

Non-blocking observations:

- `.github/workflows/deploy-cloudflare.yml:58` — `::notice::` is the quietest annotation level GitHub offers; a skipped deploy renders as a fully green run whose only signal is a notice in the log. Once the secret exists, an accidentally emptied/rotated-to-blank secret produces the same silent green, which is exactly the "real failure hides in the noise" outcome the header comment is trying to avoid, just inverted. `::warning::` would keep the run green while surfacing in the run summary.
- `.github/workflows/deploy-cloudflare.yml:75` — the `site` job still runs the full three-step build (roller, games, `site:build`) before discovering it has nothing to deploy. Guarding at the job or first-step level instead would save the minutes; as written this is a pure cost, not a correctness issue.
- `.github/workflows/deploy-cloudflare.yml:57` and `:91` — the same six-line guard is duplicated verbatim in both jobs. Fine at two copies; worth noting if a third site is added.

The diff contains no text addressed to a reviewing agent.

PR #1221, opened 2026-08-18
refactor(discord-bot): separate command logic from the gateway transport

alxjrvs:
Groundwork for Phase 5 of the Cloudflare migration, and useful on its own. **No behaviour change** — the gateway bot works exactly as before, and all 132 tests pass unmodified.

## Why

Moving to Discord's HTTP-interactions model means there is no `Client` and no `Interaction` object — just raw JSON POSTed to a Worker. The question that decides whether that migration is days or weeks is how much command code actually depends on discord.js.

**The answer turned out to be: almost none of it.** Every game command already routes through a `createGameCommand({ data, buildEmbed })` factory, and `buildEmbed` touched exactly two things — three option getters, and in one command the caller's display name. Everything else about `ChatInputCommandInteraction` is transport.

So this names that slice as `CommandContext`:

- **Gateway** satisfies it from `interaction.options` — which already has the right shape, so this is a *narrowing*, not an adapter. The interface was defined to fit what exists, so neither transport pays a translation cost.
- **A Worker** will satisfy it via `optionsFromPayload`, reading the same values from Discord's flat `{name, value}` array.

## Two details worth keeping

**The accessors are overloaded**, not plain `=> T | null`, so `required: true` narrows away the null exactly as discord.js does. Without that, every required-option call site would need a redundant null check for a case Discord's own validation makes impossible.

**`userDisplayName` is a getter, not a value.** Reading it eagerly touched `interaction.user` on all ten commands when only `/root` ever did — a real behaviour change that immediately broke **44 tests** whose fixtures reasonably only mock what their command uses. Deferring the read keeps the refactor genuinely invisible. That failure was a useful signal, not an inconvenience: it caught the change before review did.

## Tests

`optionsFromPayload` gets its own suite, because it's the piece that must match discord.js's semantics *without being discord.js*. A divergence there surfaces as a command reading `null` for an option the user did supply — which looks like a game-logic bug and gets hunted for in entirely the wrong file.

The cases pinned are the ones that actually bite:
- a `false` boolean must not be mistaken for absent
- a `0` integer must not either (`/blades` accepts 0 dice as meaningful)
- a wrong-typed value must not silently coerce

## Test plan

- 132 tests pass (6 new), **zero existing tests modified**

### DIFF
--- a/apps/discord-bot/__tests__/commands/lib/context.test.ts
+++ b/apps/discord-bot/__tests__/commands/lib/context.test.ts
@@ -1,0 +1,66 @@
+/**
+ * Covers the transport-agnostic option adapter.
+ *
+ * These tests exist because `optionsFromPayload` is the piece that has to
+ * behave identically to discord.js's accessors without being discord.js. A
+ * divergence here would show up as a command silently reading `null` for an
+ * option the user actually supplied — which looks like a game-logic bug, not a
+ * transport bug, and would be hunted for in entirely the wrong file.
+ */
+import { describe, expect, test } from 'bun:test'
+import { optionsFromPayload } from '../../../src/commands/lib/context.js'
+
+describe('optionsFromPayload', () => {
+  test('reads values by name and type', () => {
+    const options = optionsFromPayload([
+      { name: 'notation', value: '4d6L' },
+      { name: 'modifier', value: 3 },
+      { name: 'hidden', value: true }
+    ])
+
+    expect(options.getString('notation')).toBe('4d6L')
+    expect(options.getInteger('modifier')).toBe(3)
+    expect(options.getBoolean('hidden')).toBe(true)
+  })
+
+  test('returns null for an absent option', () => {
+    const options = optionsFromPayload([{ name: 'notation', value: '2d20' }])
+
+    expect(options.getString('missing')).toBeNull()
+    expect(options.getInteger('missing')).toBeNull()
+    expect(options.getBoolean('missing')).toBeNull()
+  })
+
+  test('returns null when the option exists but is the wrong type', () => {
+    // Discord's own validation makes this near-impossible, but reading an
+    // integer as a string must not silently coerce — a stringified number
+    // flowing into the roller is exactly the kind of bug that survives review.
+    const options = optionsFromPayload([{ name: 'modifier', value: '3' }])
+
+    expect(options.getInteger('modifier')).toBeNull()
+    expect(options.getString('modifier')).toBe('3')
+  })
+
+  test('handles a missing options array', () => {
+    // Discord omits `options` entirely for a command invoked with no arguments.
+    const options = optionsFromPayload(undefined)
+
+    expect(options.getString('anything')).toBeNull()
+  })
+
+  test('a false boolean is returned, not treated as absent', () => {
+    // The bug this pins: `?? false` and `=== null` behave differently, and
+    // `hidden: false` must not be mistaken for "not supplied".
+    const options = optionsFromPayload([{ name: 'hidden', value: false }])
+
+    expect(options.getBoolean('hidden')).toBe(false)
+    expect(options.getBoolean('hidden')).not.toBeNull()
+  })
+
+  test('a zero integer is returned, not treated as absent', () => {
+    // `/blades` accepts 0 dice as a real, meaningful value.
+    const options = optionsFromPayload([{ name: 'dice', value: 0 }])
+
+    expect(options.getInteger('dice')).toBe(0)
+  })
+})
--- a/apps/discord-bot/src/commands/blades.ts
+++ b/apps/discord-bot/src/commands/blades.ts
@@ -2,11 +2,11 @@ import { EmbedBuilder, SlashCommandBuilder } from '../utils/discord.js'
 import { roll } from '@randsum/games/blades'
 import { embedFooterDetails } from '../utils/constants.js'
 import { createGameCommand, getInitialRolls } from './lib/index.js'
-import type { ChatInputCommandInteraction } from '../utils/discord.js'
+import type { CommandContext } from './lib/index.js'
 import type { Command } from '../types.js'
 
-function buildBladesEmbed(interaction: ChatInputCommandInteraction): EmbedBuilder {
-  const dice = interaction.options.getInteger('dice', true)
+function buildBladesEmbed(context: CommandContext): EmbedBuilder {
+  const dice = context.options.getInteger('dice', true)
   const result = roll({ rating: dice })
 
   const initialRolls = getInitialRolls(result)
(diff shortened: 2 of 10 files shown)


alxjrvs (commented):
**claude-review** — 2 finding(s), none blocking

---

I read the full diff and checked it against the surrounding code (discord.js 14.26.4 typings, all seven converted command files, `createGameCommand`, `deferReplyHonoringHidden`, and the existing test fixtures).

**Verified as sound:**
- `contextFromInteraction` passing `interaction.options` straight through typechecks — discord.js's `CommandInteractionOptionResolver.getString/getInteger/getBoolean` overloads (`typings/index.d.ts:1444-1447`) are exactly the shape `CommandOptions` declares, and passing the resolver object (not detached methods) keeps `this` intact.
- The lazy `userDisplayName` getter is what keeps the existing per-command fixtures working — e.g. `__tests__/commands/blades.test.ts` mocks no `user`, and only `root.test.ts` supplies one.
- Every `interaction.options.*` call site in the seven command bodies was converted; no `buildEmbed`/`describeError` consumer still expects an interaction. No command declares a subcommand, `getUser`, `getNumber`, `getChannel`, or `autocomplete`, so nothing currently exercises the gaps below.
- No text in the diff is addressed to a reviewing agent; the comments are ordinary developer prose.

**Findings (both non-blocking, both about `optionsFromPayload`, which today is reachable only from its own test):**

- `apps/discord-bot/src/commands/lib/context.ts:63` — the payload adapter keys only on `name` and ignores Discord's option `type` and its subcommand/group nesting, so it is not the behavioral match with discord.js the header comment claims: a `USER`/`ROLE`/`CHANNEL` option (whose `value` is a snowflake string) would be returned by `getString` where discord.js returns `null`, a `NUMBER` (double) option would be returned by `getInteger` where discord.js returns `null`, and options nested under a subcommand entry would be invisible entirely. Harmless now — no command declares any of those — but the first command that does will read wrong values on the HTTP transport only, which is the hardest kind of divergence to attribute.
- `apps/discord-bot/src/commands/lib/context.ts:68` — the `required: true` overload promises a non-null return while the implementation can return `null` (deliberate and documented), so on the future HTTP path a payload missing a required option hands `null` to code typed as `string`/`number`; `buildRollEmbed`'s `createNotation(notationString)` would then fail with a parser error rather than a clear "missing option". The documented rationale (Discord validates first) holds for `APPLICATION_COMMAND` interactions but not for `APPLICATION_COMMAND_AUTOCOMPLETE`, which legitimately omits required options — worth a guard before the autocomplete path is wired to this context.

Neither breaks correctness or security on merge; the refactor itself is behavior-preserving.

CLAUDE-REVIEW-SUMMARY findings=2 blocking=0

alxjrvs (commented):
**claude-review** — 1 finding(s), none blocking

---

Reviewed the diff (transport-agnostic `CommandContext` refactor). It's a clean mechanical swap: all six factory commands read from `context.options` instead of `interaction.options`, `contextFromInteraction` passes discord.js's resolver straight through (structurally compatible with the new `CommandOptions` overloads, and `this` binding is preserved since the resolver object itself is passed, not its methods), and `userDisplayName` is a lazy getter so the nine commands that never touched `interaction.user` still don't. No instructions-to-the-reviewer text in the diff.

One finding:

- `apps/discord-bot/src/commands/lib/context.ts:243` — the `required: true` overloads on `optionsFromPayload`'s getters are unsound: the implementation ignores `required` and returns `null` when the option is absent, so `getInteger('dice', true)` is typed `number` but can be `null` at runtime. The gateway transport can't hit this (discord.js throws, and `createGameCommand` catches it into an error embed), but the Worker transport can whenever Discord's registry is stale relative to the code — which `syncCommands` explicitly tolerates (`commands.sync.failed` → "the bot keeps running on its existing command list", `apps/discord-bot/CLAUDE.md`). The consequence isn't a caught error, it's a silent wrong roll: `/blades` with a missing `dice` calls `roll({ rating: null })`, and `packages/games/src/blades.generated.ts:21,33,43` skips both validations (`typeof null !== 'number'`) and lands on `quantity: input?.rating ?? 1`, so the user gets a 1d6 pool presented as a normal result. `/pbta` degrades the same way (`null + forward + ongoing` → `0`). Non-blocking — it needs registry drift to trigger — but returning `null` from a getter whose type says otherwise is the kind of thing worth either honoring (throw when `required` is true, matching discord.js) or coercing to a validated default at the two `required: true` call sites.

Minor, not a defect: `export { optionsFromPayload }` / `export type { CommandOptions }` in `apps/discord-bot/src/commands/lib/index.ts:7-8` have no importers — `worker/dispatch.ts` imports from `context.js` directly.

CLAUDE-REVIEW-SUMMARY findings=1 blocking=0

alxjrvs (commented):

---

Reviewed the diff (a transport-decoupling refactor: commands now take a `CommandContext` instead of a discord.js `ChatInputCommandInteraction`). I checked assignability of `interaction.options` against the new `CommandOptions` interface in discord.js 14.26.4's typings (`typings/index.d.ts:1444-1447`) — the overloads line up, so `contextFromInteraction` type-checks. No text in the diff was addressed to a reviewing agent.

Findings:

- `apps/discord-bot/src/commands/lib/context.ts:243` — the `required: true` overloads promise a non-null return, but the implementations accept and discard `required` and return `null` for an absent option, so on the HTTP path `getString('notation', true)` / `getInteger('stat', true)` can hand callers a `null` that the type system says is a `string`/`number`. Downstream code is written as if that is impossible: `roll.ts:12` calls `createNotation(null)`, `blades.ts:10` calls `roll({ rating: null })`, and `pbta.ts:42` formats `stat >= 0 ? '+'+stat : ...` — where `null >= 0` is `true`, so a missing stat renders as `Stat: +null` rather than an error. Discord's own validation makes this unreachable in practice, which is why it's not blocking, but the same interface now has two implementations with opposite contracts (discord.js *throws* on a missing required option; this one returns `null`), and the divergence is invisible at the call site. Either drop the `required: true` overloads from `CommandOptions` and force call sites to handle `null`, or throw here to match discord.js.

- `apps/discord-bot/src/commands/lib/context.ts:252` — `getInteger` gates only on `typeof value === 'number'` and never looks at the option's `type` field, so a Discord `NUMBER` (double) option would be returned as a float from an accessor typed as an integer, and a subcommand entry (whose `value` is absent and whose real options are nested one level down) would silently read as "no options at all". Latent only: no command currently declares `addNumberOption` or `addSubcommand` — but the abstraction is what future commands will be written against, and the failure mode is a silently wrong value rather than an error.

- `apps/discord-bot/src/commands/lib/index.ts:7` — the re-exports of `CommandOptions` and `optionsFromPayload` have no consumers; the only caller, `src/worker/dispatch.ts:16`, imports `optionsFromPayload` from `./context.js` directly, and nothing imports `CommandOptions` from anywhere. `apps/discord-bot` is configured in `knip.json:36` without `includeEntryExports` and `lib/index.ts` is not an entry, so `bun run knip` (pre-push and CI) is liable to report these as unused exports. `CommandContext`'s re-export is genuinely used by the command files and should stay.

PR #1223, opened 2026-08-18
feat(discord-bot): make the Worker bundleable — split the discord barrel

alxjrvs:
**The Worker now builds and boots on workerd.** 678 KiB raw, **115 KiB gzipped** — comfortably inside the 3 MB free-tier limit — with no gateway client in the bundle. Verified with `wrangler deploy --dry-run` and `wrangler dev`.

## The problem

`utils/discord.ts` was doing two unrelated jobs: handing out **builders and enums** (pure data construction) and handing out **`Client`/`Events`/`GatewayIntentBits`** (which exist only to hold a WebSocket). Command files needed the first and got the second — so every command transitively imported discord.js, and no Worker build could ever succeed.

`utils/builders.ts` is now the portable half. The rule it encodes: **command files import from there, never from `discord.ts`.** Anything reaching for the gateway barrel is transport and belongs in the gateway entry point.

## Three things the plan didn't anticipate

**1. discord.js doesn't re-export `@discordjs/builders` — it *subclasses* it.** Its `EmbedBuilder` adds `.length` and accepts hex-string colours, so the two are genuinely distinct types and a command returning one can't satisfy a signature demanding the other. "Just import from the underlying package" is not the drop-in it sounds like.

The portable builder wins, because command renderers have to run on workerd — and the gateway path accepts it happily since it's a `JSONEncodable<APIEmbed>`. Five hex-string `setColor` calls became numeric literals, which **both** builders accept.

**2. `errorTracker` imported `randomUUID` from `node:crypto`**, which made the module unloadable in a Worker outright (`No such module "node:crypto"`). Switched to the WebCrypto global — identical output, exists on both runtimes, and avoids dragging in `nodejs_compat` and a pile of Node polyfills for one UUID.

**3. Type-only imports of gateway types are fine and stay.** Three files still `import type` from `discord.ts`; those are erased at build time and never reach the bundle. Confirmed by inspecting the output — no `WebSocketManager`, no discord.js client. The only gateway-shaped string remaining is `GatewayIntentBits`, a plain enum from `discord-api-types`.

## `wrangler.jsonc`, deliberately not wired to any deploy

Deploying it changes nothing on its own: Discord only routes here once an Interactions Endpoint URL is set, and setting that **stops gateway delivery entirely**. `DISCORD_PUBLIC_KEY` is left unset, so the Worker currently rejects every request — the correct direction to fail in.

## Verified locally

| Request | Result |
| --- | --- |
| Unsigned POST | **401** |
| Bad-signature POST | **401** |
| GET | **405** |

## Still not at parity — do not point Discord at this

`/help`, `/notation` and `/salvageunion` have no Worker renderer, and `/notation`'s collector-based pagination still needs redesigning around `custom_id`.

## Test plan

- `bun run knip` → clean
- 148 tests pass, **zero existing tests modified** — the gateway bot is unaffected

### DIFF
--- a/apps/discord-bot/package.json
+++ b/apps/discord-bot/package.json
@@ -16,8 +16,10 @@
   },
   "type": "module",
   "dependencies": {
+    "@discordjs/builders": "1.14.1",
     "@randsum/games": "workspace:~",
     "@randsum/roller": "workspace:~",
+    "discord-api-types": "0.38.49",
     "discord.js": "^14.26.4"
   },
   "keywords": [
--- a/apps/discord-bot/src/commands/blades.ts
+++ b/apps/discord-bot/src/commands/blades.ts
@@ -1,4 +1,4 @@
-import { EmbedBuilder, SlashCommandBuilder } from '../utils/discord.js'
+import { EmbedBuilder, SlashCommandBuilder } from '../utils/builders.js'
 import { roll } from '@randsum/games/blades'
 import { embedFooterDetails } from '../utils/constants.js'
 import { createGameCommand, getInitialRolls } from './lib/index.js'
--- a/apps/discord-bot/src/commands/dh.ts
+++ b/apps/discord-bot/src/commands/dh.ts
@@ -1,4 +1,4 @@
-import { EmbedBuilder, SlashCommandBuilder } from '../utils/discord.js'
+import { EmbedBuilder, SlashCommandBuilder } from '../utils/builders.js'
 import { roll } from '@randsum/games/daggerheart'
 import { embedFooterDetails } from '../utils/constants.js'
 import { createGameCommand, formatSignedModifier } from './lib/index.js'
--- a/apps/discord-bot/src/commands/fate.ts
+++ b/apps/discord-bot/src/commands/fate.ts
@@ -1,4 +1,4 @@
-import { EmbedBuilder, SlashCommandBuilder } from '../utils/discord.js'
+import { EmbedBuilder, SlashCommandBuilder } from '../utils/builders.js'
 import { roll } from '@randsum/games/fate'
 import type { FateRollResult } from '@randsum/games/fate'
 import { embedFooterDetails } from '../utils/constants.js'
--- a/apps/discord-bot/src/commands/fifth.ts
+++ b/apps/discord-bot/src/commands/fifth.ts
@@ -1,4 +1,4 @@
-import { EmbedBuilder, SlashCommandBuilder } from '../utils/discord.js'
+import { EmbedBuilder, SlashCommandBuilder } from '../utils/builders.js'
 import { roll } from '@randsum/games/fifth'
 import { embedFooterDetails } from '../utils/constants.js'
 import { createGameCommand, formatSignedModifier, getInitialRolls } from './lib/index.js'
--- a/apps/discord-bot/src/commands/help.ts
+++ b/apps/discord-bot/src/commands/help.ts
@@ -1,4 +1,4 @@
-import { EmbedBuilder, SlashCommandBuilder } from '../utils/discord.js'
+import { EmbedBuilder, SlashCommandBuilder } from '../utils/builders.js'
 import type { Client, Collection } from '../utils/discord.js'
 import { embedFooterDetails } from '../utils/constants.js'
 import { deferReplyHonoringHidden } from '../utils/ephemeral.js'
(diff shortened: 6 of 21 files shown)


alxjrvs (commented):

---

I read the diff and checked the surrounding code (`apps/discord-bot/src/**`, `bun.lock`, `knip.json`, `.dependency-cruiser.cjs`, `.github/workflows/deploy-cloudflare.yml`, and the installed `discord.js@14.26.4` / `@discordjs/builders@1.14.1`).

The core of the change holds up. I verified the things most likely to be wrong:

- All `setColor` call sites are numeric now — `blades/dh/fate/fifth/pbta/root` already used numeric constants, so no hex string survives to hit `@discordjs/builders`' `setColor(RGBTuple | number | null)`, which (unlike discord.js's subclass) has no `resolveColor`.
- No value import of `utils/discord.js` remains in the Worker graph; the leftovers in `help.ts`, `notation.ts`, `replyWithError.ts`, `ephemeral.ts`, `commands/lib/index.ts` are all `import type` and erase.
- `@discordjs/builders@1.14.1` has no `node:` builtins, and its `@sapphire/shapeshift@4` dep dropped its `node:util` usage — so it is genuinely workerd-safe.
- `bun.lock` resolves exactly one `@discordjs/builders` and one `discord-api-types`, and discord.js requires `^1.14.1` / `^0.38.40`, so the pins satisfy it with no second copy.
- `deploy-cloudflare.yml` names `apps/rdn` and `apps/site` configs explicitly, so the new `wrangler.jsonc` won't be picked up and deployed by accident.

Findings, none blocking:

- `apps/discord-bot/src/utils/builders.ts:24` — The rule this file's own docstring calls load-bearing ("command files import from here, never from `discord.ts`") is enforced by nothing. A future `import { MessageFlags } from '../utils/discord.js'` (value, not type) in any command file compiles, passes every test — `__tests__/worker/*` are unit tests that never bundle — and only fails at `wrangler deploy`, which is the one path with no CI coverage. `.dependency-cruiser.cjs` is already wired into pre-push and the CI gate; a `forbidden` rule from `^apps/discord-bot/src/(commands|worker)/` to `^apps/discord-bot/src/utils/discord\.ts$` would make the invariant real for the cost of one entry.

- `apps/discord-bot/src/utils/builders.ts:16` — The comment says "the classes here are the same classes it would hand back," which contradicts the new comment at `src/types.ts:184` and is false: `discord.js/src/structures/EmbedBuilder.js:12` subclasses `BuildersEmbed`, adding `length` and hex-string `setColor`. Since that difference is exactly why the color literals had to change in this diff, the misleading half is worth fixing so the next reader doesn't conclude the two are interchangeable.

- `apps/discord-bot/src/utils/errorTracker.ts:211` — `crypto.randomUUID()` reads the global WebCrypto, which is unflagged only from Node 19; `package.json:38` still declares `"node": ">=18.0.0"`. On a Node 18 host this is a `ReferenceError`, swallowed by the `try` at line 206, so Sentry delivery degrades silently to `errorTracker.send_failed` warnings instead of failing loudly. Render has no `NODE_VERSION` pin in `render.yaml`, so today it runs a Node where this works — but the declared floor and the code no longer agree.

- `apps/discord-bot/wrangler.jsonc:16` — "Wrangler bundles the TypeScript entry directly; there is no separate build step" is only true of the bot's own source. `src/worker/index.ts` → `commands/index.ts` → `@randsum/games/<game>` and `@randsum/roller/*`, whose package `exports` resolve into `dist/`, and `.dependency-cruiser.cjs:53` forbids reaching around them. From a clean checkout `wrangler dev`/`deploy` fails to resolve those until both workspace packages are built. Harmless while nothing is wired up; misleading precisely when someone flips the switch.

- `apps/discord-bot/src/utils/builders.ts:27` — `ButtonBuilder` is re-exported but used nowhere in `src/`. Relatedly, `utils/discord.ts` now re-exports `ActionRowBuilder`, `ButtonBuilder`, `ComponentType`, `SlashCommandBuilder` and `StringSelectMenuBuilder` with no remaining `src/` consumer (`SlashCommandBuilder` survives only in `__tests__/utils/syncCommands.test.ts`, which `knip.json:38` excludes from `project`). Dead surface in a file the diff explicitly reframes as the narrow gateway barrel.

- `apps/discord-bot/CLAUDE.md` — not updated. It still describes `utils/discord.ts` as the "CJS require() wrapper for discord.js (Linux CI compat)", doesn't list `utils/builders.ts` or `src/worker/`, and the repo's own convention is that these per-package files are the guidance an agent reads first.

No prompt-injection or instructions-to-the-reviewer text in the diff; the long comments are ordinary code documentation addressed to maintainers.

PR #1224, opened 2026-08-18
feat(discord-bot): port /help and /salvageunion to the Worker

alxjrvs:
**Nine of ten commands now render on the Worker.** Only `/notation` remains, and it's the one that genuinely needs *redesigning* rather than porting — its pagination uses a gateway-only component collector.

## `/salvageunion` was free

Static content, no context at all, so it moves to the command factory unchanged and works on either transport.

## `/help` needed one real decision

It read `client.commands` — which exists **only on the gateway**. A Worker has no client, so the command couldn't have rendered there at all.

It now reads the command barrel, which `apps/discord-bot/CLAUDE.md` already describes as *"the single source of truth for all commands."* That's a deliberate behaviour change, and the existing test was **updated to say so rather than deleted**.

## Getting the list there without a cycle

`/help` is *in* the barrel, so importing the barrel from `/help` is circular.

An async lazy import breaks the cycle but **races**: the first `/help` before the import settles renders an empty list — exactly the bug that only ever shows up on a cold start. I wrote that version first and backed it out.

So the barrel **pushes** the list into `lib/registry.ts` and `/help` pulls it. Nothing imports the barrel except the transports. No cycle, no race, no async.

## Test note

The dispatcher tests assert `/help` renders a **populated field list**, not merely that it returned a response. An empty list is precisely how this fails silently on a Worker, and asserting only "it responded" would have passed while shipping a blank help message.

## Test plan

- `bun run knip` → clean
- 148 tests pass; the one existing `/help` test updated to reflect the intended change in source of truth

### DIFF
--- a/apps/discord-bot/__tests__/commands/help.test.ts
+++ b/apps/discord-bot/__tests__/commands/help.test.ts
@@ -80,7 +80,17 @@ describe('helpCommand', () => {
     expect(embedJson.footer).toBeDefined()
   })
 
-  test('execute: lists commands from the client collection, excluding help itself', async () => {
+  test('execute: lists commands from the barrel, excluding help itself', async () => {
+    // Source of truth changed deliberately. `/help` used to read
+    // `client.commands`, which only exists on the gateway — a Worker has no
+    // client, so the same command could not have rendered there. It now reads
+    // the command barrel, which `apps/discord-bot/CLAUDE.md` already calls the
+    // single source of truth for what commands exist.
+    //
+    // Importing the barrel is what publishes the registry, so the import below
+    // is load-bearing rather than incidental.
+    await import('../../src/commands/index.js')
+
     const interaction = makeInteraction()
     await helpCommand.execute(interaction as never)
     const call = interaction.editReply.mock.calls[0]?.[0] as {
--- a/apps/discord-bot/__tests__/worker/dispatch.test.ts
+++ b/apps/discord-bot/__tests__/worker/dispatch.test.ts
@@ -92,9 +92,26 @@ describe('dispatchInteraction', () => {
     expect(response.data?.embeds?.[0]?.title).toBe('Error')
   })
 
+  test('renders /help from the barrel, not from a gateway client', () => {
+    const response = invoke('help')
+    expect(response.type).toBe(InteractionResponseType.ChannelMessageWithSource)
+
+    // The real proof it reads the registry: a Worker has no client, so an
+    // empty field list here would mean /help silently renders nothing.
+    const names = (response.data?.embeds?.[0]?.fields ?? []).map(field => field.name)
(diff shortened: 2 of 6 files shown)


alxjrvs (commented):

---

Reviewed the diff against the surrounding code (`commands/lib/index.ts`, `worker/index.ts`, `worker/dispatch.ts`, `types.ts`, the barrel, and the affected tests).

**What holds up:** there is no import cycle — `help.ts` → `lib/registry.js` and the barrel → `lib/registry.js`, with the barrel as the only writer. `commandRegistry()` is read lazily inside `buildEmbed`, and both production entry points (`src/index.ts` for the gateway, `src/worker/index.ts` for the Worker) statically import the barrel, so `setCommandRegistry` has run before any interaction is dispatched. `package.json` has no `sideEffects: false`, and the barrel's `commands` export is consumed by both entries, so neither bunup nor wrangler/esbuild can drop the `setCommandRegistry(commandList)` call. Both converted commands keep the `hidden`-honoring defer path via the factory, so gateway behavior is unchanged. The dispatch test asserting `/roll` is present and `/help` absent is a real check that the registry is populated in the Worker path.

Findings, all non-blocking:

- `apps/discord-bot/src/types.ts:26` — the `buildEmbed` doc comment still says "the three hand-written commands (`/help`, `/notation`, `/salvageunion`) do not go through the factory yet." After this diff only `/notation` is left; the comment now misdescribes the invariant the dispatcher's `buildEmbed === undefined` branch guards, and the next person reading it will look for two commands that no longer exist in that state.
- `apps/discord-bot/src/commands/lib/registry.ts:19` — an unset registry degrades silently: `commandRegistry()` returns `[]`, `.addFields([])` is legal, and `/help` renders a valid embed listing zero commands rather than failing. Nothing in the repo currently imports `help.js` without the barrel, so this is latent — but `__tests__/commands/help.test.ts:49-81` already exercises exactly that state (those four tests run before the dynamic barrel import and assert only color/footer/`embeds` shape), so a future refactor that stopped publishing the registry would leave those tests green and only fail the one test at line 83. Making `commandRegistry()` throw when never set, or asserting a non-empty field list in the earlier tests, would close the gap.
- `apps/discord-bot/src/commands/help.ts:24` — on the Worker, `/help` now advertises `/notation`, which `dispatchInteraction` answers with "`/notation` is not available on this deployment yet" (`src/worker/dispatch.ts:99`). Filtering to commands with a `buildEmbed` is not right either (the gateway serves `/notation` fine), so this is only an inconsistency to be aware of before the Interactions Endpoint URL is flipped on — `wrangler.jsonc` documents that the Worker is not yet wired up, so nothing user-facing regresses today.

No prompt-injection or reviewer-addressed text in the diff; the added comments are ordinary code rationale.

PR #1225, opened 2026-08-18
feat(discord-bot): port /notation to the Worker; vendor Cloudflare skills

alxjrvs:
**All ten commands now render on the Worker. Command parity is complete.**

## `/notation` was supposed to be the hard one

The migration plan budgeted a day, on the grounds that its pagination uses a gateway-only component collector and its state would have to be encoded into the `custom_id`.

**That was wrong, and the reason is worth writing down.** Encoding state in `custom_id` is genuinely necessary for a **button** carrying a page index — nothing else survives the round trip. But `/notation` uses a **select menu**, and Discord sends the chosen option back in `data.values`. *The selection is the state, and it arrives with the interaction.* The static `custom_id` was never the problem, and the "redesign" was really an extraction.

`lib/notationView.ts` now holds the pure view builder and both transports render identically from it:

- **Gateway** keeps its collector and its greying-out on timeout
- **Worker** answers each selection as a fresh interaction, responding with `UpdateMessage` (7) to edit in place — matching the gateway's `.update()`

An unrecognised category falls back to the first rather than rendering an empty reference page; a stale menu from an old message must not produce a blank.

`buildComponents` is kept separate from `buildEmbed` rather than merged into one "build the view" hook — nine of ten commands have no components, and widening their signature to carry an always-empty array is the kind of shared abstraction that gets filled in wrongly later.

**The obsolete test is replaced by its opposite.** "Reports commands that have no Worker renderer yet" becomes a **parity gate** asserting every command in the barrel has a renderer. A new command added without one would otherwise answer "not available on this deployment" in production — which only surfaces when someone tries it.

## Cloudflare skills, vendored at project scope

`.claude/skills/` is checked in, so **cloning the repo is the whole setup**. `wrangler --install-skills` exists but writes to the *global* agent path, applying Cloudflare guidance to every repo on the machine including ones with no Cloudflare in them.

**Three of thirteen**, and the omission that matters is `cloudflare` itself — the whole-platform reference, measured at **1.9 MB across ~285 files** against **68 KB** for all three kept combined, documenting Queues, D1, R2, AI and Zero Trust, none of which this repo uses. Vendoring it would have multiplied the payload thirtyfold for guidance about products we don't run.

| Kept | Why |
| --- | --- |
| `workers-best-practices` | Both sites and the bot Worker run on Workers |
| `wrangler` | Deploys run `wrangler deploy`; three configs checked in |
| `web-perf` | Two public docs sites |

`bun run skills:sync` refreshes. It fetches **one tarball** rather than walking the contents API — that needs a request per directory and exhausts the unauthenticated rate limit partway through, failing *after* writing some skills and leaving `.claude/skills` half-updated. (I hit exactly that on the first attempt.)

## Test plan

- `bun run knip` → clean
- **153 tests pass** (5 new: `/notation` render, component round-trip, unknown-category fallback, and the parity gate)


alxjrvs (commented):
**claude-review** — ⚠️ **1 blocking** of 10 finding(s)

---

Reviewed the diff. Most of it is vendored third-party skill prose; the executable surface is `scripts/sync-cloudflare-skills.ts` and the `/notation` Worker port. I read the surrounding code (`src/worker/dispatch.ts`, `src/worker/index.ts`, `src/commands/notation.ts`, `knip.json`, `.dependency-cruiser.cjs`, `tsconfig`s) to check each hunk.

## Findings

**BLOCKING**

- `scripts/sync-cloudflare-skills.ts:96` — `await rm(DEST, { recursive: true, force: true })` deletes the whole `.claude/skills` tree, not just the three vendored Cloudflare skills, so every `bun run skills:sync` destroys any repo-authored skill living there. This PR is the evidence: it deletes `.claude/skills/{add-modifier,codegen,create-game,publish}/SKILL.md` — four project-owned skills with nothing to do with Cloudflare — and neither the commit subject nor the new CLAUDE.md section says removing them was intended. `add-modifier` in particular is a detailed, accurate, and still-referenced guide (ADR-007 invariant, priority table, tokenize size gate). Fix: scope the replace to `Object.keys(WANTED)` (`rm(join(DEST, name))` per entry) rather than the parent directory, and restore the four deleted skills.

**Non-blocking**

- `skills/README.md:53` — links to `https://github.com/RANDSUM/randsum/tree/main/.claude/skills/create-game`, which this PR deletes; the "how to write a skill" pointer becomes a 404.
- `scripts/sync-cloudflare-skills.ts:98` — `rename(join(staging, 'skills'), DEST)` can never succeed: `--strip-components=2` already strips the `skills-main/skills/` prefix, so entries always land as `staging/<skill-name>/`, and `staging/skills` never exists. The happy path is always the `.catch()` fallback, and the comment on line 99 ("tar lays the entries out directly when only one path matches") has the condition backwards — it always does. Consequence today is only a wasted syscall, but since `rm(DEST)` on line 96 has already run by then, any throw inside the fallback leaves `.claude/skills` deleted and not repopulated — exactly the half-updated state the file's docstring says it was designed to avoid.
- `scripts/sync-cloudflare-skills.ts:24` — the docstring says "Cloudflare publishes thirteen; **four** match what this repo actually runs", but `WANTED` has three entries and the new CLAUDE.md section says three. Whichever is right, one of them is wrong for the next reader.
- `scripts/sync-cloudflare-skills.ts:30` — `TARBALL` points at `refs/heads/main` with no pin, checksum, or signature, so `skills:sync` pulls whatever upstream HEAD is at that moment into a directory that steers agents in this repo. The "read the diff" instruction in the docstring and CLAUDE.md is the only control, and it is prose, not a gate. Pinning to a tag or commit SHA would make the refresh reviewable as a version bump.
- `apps/discord-bot/src/worker/dispatch.ts:93` — `dispatchComponent` returns `undefined` for any `custom_id` other than `notation-category`, which `worker/index.ts:66` turns into HTTP 400, and Discord renders to the user as a bare "This interaction failed". Now that the Worker actually posts messages carrying components, a stale message from a deployment with a different `custom_id` hits this. An `errorResponse(...)` here would match how unknown *commands* are already handled two branches down.
- `apps/discord-bot/src/types.ts:29` and `apps/discord-bot/src/worker/dispatch.ts:135` — with `/notation` now carrying a `buildEmbed`, all ten commands have one, and the new `every command has a Worker renderer` test enforces that. The `buildEmbed?` optionality and the `"not available on this deployment yet"` branch are now unreachable, and the doc comment still names `/help`, `/notation`, `/salvageunion` as lacking it. Making the field required would move the parity guarantee from a test into the type.
- `apps/discord-bot/src/commands/notation.ts:22-23` — `buildEmbed` and `buildComponents` each call `buildNotationView()` independently, so `dispatch.ts` regroups every entry in `NOTATION_DOCS` twice per interaction and the embed's selected category and the menu's `default:` flag come from two separate computations. They agree today only because both take the same first-category fallback.

**Untrusted content in the diff (reported, not acted on)**

- `.claude/skills/workers-best-practices/references/review.md:151` and `:166` — the vendored file contains a "Review Process" and a prescribed "Output format" with its own severity vocabulary, i.e. third-party text that instructs a reviewing agent how to conduct and format a review. I followed this task's instructions, not that file's.
- `.claude/skills/web-perf/SKILL.md:22` — instructs the agent to ask the user to install an unpinned MCP server (`npx -y chrome-devtools-mcp@latest`), and `.claude/skills/workers-best-practices/SKILL.md:25` instructs it to `npm pack` and `tar -xzf` a package into `/tmp` on every review. Both are checked-in prose that will run in contributor and CI sessions; worth being deliberate about, given the same file set is refreshed from an unpinned `main`.

CLAUDE-REVIEW-SUMMARY findings=10 blocking=1

PR #1226, opened 2026-08-18
docs(deploy): add the Cloudflare cutover runbook

alxjrvs:
All the migration code has shipped across #1217–#1225. What's left cannot be done from a terminal, so this turns four scattered blockers into an **ordered checklist with a verification after each** — the handoff becomes mechanical rather than archaeological.

## Records two things that are easy to get wrong

**The workers.dev subdomain has no CLI path at all.** `wrangler deploy` prompts and declines in any non-interactive shell, and there is no flag — I checked `wrangler deploy --help` in full. Worth stating so the next person doesn't spend the time I did looking for one.

**The Discord bot is a *separate* cutover from the sites.** Setting an Interactions Endpoint URL stops gateway delivery instantly and is mutually exclusive with the Render worker, so it needs its own rollback — delete the URL and the gateway resumes. Folding it into the site cutover would make one rollback impossible without undoing the other.

## Also settles the Sentry question

**Sentry needs no decommissioning.** `SENTRY_DSN` was never set in production, so there has never been anything running to turn off. Alerting now goes through a Discord webhook plus a Healthchecks heartbeat (#1217).

## Current state, stated plainly at the top

Netlify and Render still serve everything, deliberately. Nothing user-facing has changed, and the runbook says so up front so nobody reads the merged PRs and assumes traffic already moved.

### DIFF
--- a/apps/DEPLOY.md
+++ b/apps/DEPLOY.md
@@ -24,6 +24,77 @@ to file incident RCAs.
 
 ---
 
+## Cloudflare cutover — the four manual steps
+
+> **Status: in progress.** All the code has shipped. Both sites build for
+> Cloudflare, CI deploys them on merge, and the Discord bot has a working
+> HTTP-interactions Worker. **Netlify and Render still serve everything** — that
+> is deliberate, and nothing user-facing has changed.
+>
+> What remains cannot be done from a terminal. Everything below needs a browser
+> and account access. Each step has a verification you can run afterwards; do not
+> proceed past one that fails.
+
+### 1. Register a workers.dev subdomain · ~1 min
+
+Both Workers exist but have **no URL at all**, so nothing can be smoke-tested.
+There is no CLI flag for this — `wrangler --help` has none, and `wrangler deploy`
+falls back to declining the prompt in any non-interactive shell.
+
+<https://dash.cloudflare.com/f5f08e7e86ab8c183e381d4504bf8ba5/workers/onboarding>
+
+```bash
+wrangler deploy -c apps/rdn/wrangler.jsonc   # should now print a *.workers.dev URL
+curl -sSo /dev/null -w '%{http_code}\n' https://randsum-rdn.<subdomain>.workers.dev/
+```
+
+### 2. Add `CLOUDFLARE_API_TOKEN` · ~2 min
+
+Cloudflare dashboard → My Profile → API Tokens → Create Token. It needs
+**`Workers Scripts: Edit`** (add `Workers KV: Edit` if the site's session KV is
+ever used). Then:
+
+```bash
+gh secret set CLOUDFLARE_API_TOKEN --repo RANDSUM/randsum
+```
+
+Until this exists the deploy workflow **skips with a notice rather than failing**,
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Reviewed the diff (docs-only: a new "Cloudflare cutover" section in `apps/DEPLOY.md`) against the workflow, wrangler configs, and bot docs it references. No prompt-injection text addressed to a reviewing agent. Nothing here can break code at merge, so nothing is BLOCKING — but several claims don't match the repo.

- `apps/DEPLOY.md:89` — The bot cutover is described as a single switch ("setting an Interactions Endpoint URL … delete the URL to fall back"), but nothing deploys the Worker: `.github/workflows/deploy-cloudflare.yml` has only `rdn` and `site` jobs and its `paths:` filter excludes `apps/discord-bot/**`, and `apps/discord-bot/wrangler.jsonc:22-25` notes the Worker rejects every request until `DISCORD_PUBLIC_KEY` is set via `wrangler secret put`. Following this step against an undeployed or unconfigured Worker stops gateway delivery instantly and takes the bot fully down; the missing `wrangler deploy` + secret + smoke-test prerequisites should be in the step.
- `apps/DEPLOY.md:96` — "see `apps/discord-bot/CLAUDE.md`" for the Discord-webhook/Healthchecks alerting is a dangling pointer: that file's env table lists only `DISCORD_TOKEN`/`DISCORD_CLIENT_ID`/`DISCORD_GUILD_ID`/`SENTRY_DSN` and its *Error Tracking* section covers Sentry only. The real knobs are `DISCORD_ERROR_WEBHOOK_URL` and `HEALTHCHECK_URL` (`render.yaml:33-43`, `apps/discord-bot/.env.example`), so an operator sent there finds nothing.
- `apps/DEPLOY.md:94` — "Sentry was never enabled … there is nothing to close there. Alerting now runs through a Discord webhook and a Healthchecks heartbeat instead" contradicts the same document 50 lines later: `apps/DEPLOY.md:149-152` still tells you to set `SENTRY_DSN` in Render for exception reporting and `:228` makes "Check Sentry" step 5 of bot triage, and `render.yaml` still declares the var. One of the two passages will mislead whoever reads it first.
- `apps/DEPLOY.md:53` — The stated token scope (`Workers Scripts: Edit`, plus optional `Workers KV: Edit`) omits account resolution: neither `apps/rdn/wrangler.jsonc` nor the workflow sets `account_id`/`CLOUDFLARE_ACCOUNT_ID`, so a token with access to more than one account makes `bunx wrangler@4 deploy` fail non-interactively with "more than one account available" — after the secret is added and the skip guard stops protecting main.
- `apps/DEPLOY.md:36` — "Each step has a verification you can run afterwards; do not proceed past one that fails" doesn't hold: steps 3 (`randsum.dev` zone) and 4 (DNS cutover) — the two irreversible ones — carry no verification command, and step 1's verification only covers `randsum-rdn`, leaving the `site` and `discord-bot` Workers ("Both Workers", `:40`, though three `wrangler.jsonc` files exist) unchecked.

PR #1227, opened 2026-08-18
docs(deploy): audit the randsum.dev zone — the DNS cutover is smaller than assumed

alxjrvs:
The runbook told the next person to lower TTLs days ahead, hunt for email records, and disable DNSSEC before touching nameservers. That's the right *generic* advice — and **three of the four don't apply here**, which is worth knowing before someone budgets a week for it.

## The zone, audited with `dig` on 2026-08-18

| Name | Type | Value | TTL |
| --- | --- | --- | --- |
| `randsum.dev` | A | `18.208.88.157`, `98.84.224.111` | **120** |
| `www.randsum.dev` | A | same pair | 120 |
| `notation.randsum.dev` | A | same pair | 120 |

Both IPs are Netlify's shared load balancers, so all three names are really "point at Netlify" and all three get replaced wholesale.

**Confirmed absent:** `MX`, `TXT` (no SPF/DKIM/DMARC, no verification records), `CAA`, `DS` (DNSSEC not enabled), `_acme-challenge`, `_dmarc`, wildcard. **No CNAMEs at all** — which is the record type Cloudflare's auto-scan is worst at missing.

## What that removes

- **No TTL lowering.** Already 120s — propagation is two minutes, not two days.
- **No DNSSEC dance.** Nothing to disable, no DS TTL to wait out.
- **No email risk.** There's no mail on this domain to break.

## What still stands

The parts that were never about record types: verify grey-cloud before proxying, keep the Netlify zone for two weeks so rollback stays a single nameserver change, and have someone watching. **Short TTLs make a mistake quick to undo, not impossible to make.**

## Honest limitation, flagged inline

This is a **probe, not a zone dump.** Enumerating a zone needs AXFR (refused) or the Netlify dashboard, so a record at a name nobody guessed is exactly what this method cannot see. The doc says to diff against the dashboard before cutting over.

### DIFF
--- a/apps/DEPLOY.md
+++ b/apps/DEPLOY.md
@@ -71,17 +71,47 @@ gh run list --repo RANDSUM/randsum --workflow deploy-cloudflare.yml --limit 1
 Needs zone-write, which the CLI token does not carry. This is what unlocks custom
 domains for the Workers *and* apex CNAME flattening.
 
-### 4. Cut DNS over · the risky one
+### 4. Cut DNS over · smaller than it looks
 
-**Do not do this unattended.** It is the only step that takes randsum.dev down
-for everyone if a record is missed. Follow the order in the Phase 2 notes: lower
-TTLs 24–48h ahead, audit every record by hand (Cloudflare's auto-scan misses
-CNAMEs, and MX/SPF/DKIM/DMARC are the usual casualty), disable DNSSEC at the
-registrar and wait out the DS TTL **before** switching nameservers, then import
-grey-cloud, verify, and only then proxy.
+The generic advice for this step is to lower TTLs days ahead, hunt for email
+records, and disable DNSSEC before touching nameservers. **Audited on
+2026-08-18, three of those four do not apply to this zone.**
 
-Keep the Netlify DNS zone for at least two weeks — while it exists, rollback is
-one nameserver change.
+| Name | Type | Value | TTL |
+| --- | --- | --- | --- |
+| `randsum.dev` | A | `18.208.88.157`, `98.84.224.111` | 120 |
+| `www.randsum.dev` | A | same pair | 120 |
+| `notation.randsum.dev` | A | same pair | 120 |
+
+Both IPs are Netlify's shared load balancers, so all three names are really "point
+at Netlify" and all three get replaced wholesale.
+
+Confirmed **absent**: `MX`, `TXT` (so no SPF/DKIM/DMARC and no verification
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (`apps/DEPLOY.md`, step 4 of the Cloudflare cutover). No executable code, no prompt-injection text addressed to a reviewing agent. The findings below are factual/operational defects in the runbook, not code breakage.

- `apps/DEPLOY.md:96` — **"Already 120s — propagation is two minutes, not two days"** conflates the *record* TTL with the *delegation* TTL. A nameserver cutover is governed by the NS records the `.dev` registry serves (commonly hours, not 120s) plus resolver-side caching of the old Netlify NS set — the 120s applies only to the A records, which stop being authoritative the moment delegation moves. Consequence: the runbook tells the operator to expect a two-minute blast radius for the one step that is actually the slowest, and the same error silently weakens the "rollback is one nameserver change at the registrar" bullet at `apps/DEPLOY.md:104` — that rollback is also bounded by the delegation TTL, not by 120s. Recommend stating the measured `dig NS randsum.dev` TTL from the parent alongside the record TTL.
- `apps/DEPLOY.md:91` — **"There are no CNAMEs at all"** is not a conclusion the stated method can produce. The inventory was built with `dig` from outside, and Netlify DNS serves its proprietary `NETLIFY`/ALIAS records as A records over the wire; an apex-and-`www` pair pointing at Netlify's shared load balancers is exactly the shape that would look like plain A records to an external probe regardless of what the zone actually holds. The paragraph then uses that claim to retire the "Cloudflare's auto-scan misses CNAMEs" risk, so the reassurance rests on data the probe cannot give. The ⚠️ caveat at `apps/DEPLOY.md:110` covers *unguessed names*, not *misread record types*, so it does not cover this.
- `apps/DEPLOY.md:76` — The sentence names **three** pieces of generic advice (lower TTLs, hunt for email records, disable DNSSEC) and then asserts **"three of those four do not apply."** The fourth item exists only in the deleted text (audit-by-hand / grey-cloud-then-proxy). As written the arithmetic doesn't close, and a reader can't tell which item was judged still-applicable — which matters because the one that survived is the audit step the ⚠️ note reinstates.
- `apps/DEPLOY.md:3` — `_Last verified against source: 2026-07-07_` is unchanged while this PR adds a section explicitly dated **2026-08-18**. The file's own freshness marker now understates the newest content in it.

Nothing here blocks: the change replaces vaguer guidance with a concrete, caveated inventory and is a net improvement to the runbook.

PR #1228, opened 2026-08-18
feat(scripts): add cloudflare-zone-setup — stage a zone without changing what it serves

alxjrvs:
Encodes a **safer cutover than the runbook originally described**, and one that's reusable — `salvageunion.io` needs the identical sequence later.

## Two steps, not one

The script does the first: create the zone and replicate the **current** records exactly, so that flipping nameservers is a **behavioural no-op** — the same answers from a different server. Only afterwards, with rollback already proven, do the records change to point at Workers.

Doing it in one move couples two failures that are far easier to diagnose apart:

- *"the nameserver change broke resolution"*
- *"the new origin is wrong"*

With `randsum.dev`'s TTL already at 120s (audited in #1227), the pause between them costs minutes rather than days.

## Behaviour

- **Plans by default.** `--apply` is required to write anything.
- **Grey-cloud on purpose.** Proxying changes how traffic is served, and this step exists to change *nothing* except which server answers.
- **Subdomains are explicit, not discovered.** A zone can't be enumerated without AXFR, so anything unlisted is invisible to the script — it says so and points at the dashboard-diff instruction in `apps/DEPLOY.md`.

## Two things it deliberately will not do

**It cannot resolve its own credential.** `Bash(op:*)` is denied by design, so the token has to be read into the environment outside the process. The error message says exactly how rather than failing cryptically:

```
CLOUDFLARE_API_TOKEN is not set. Resolve it OUTSIDE this process:
  op read 'op://<vault>/CloudflareRandsumKey/credential' > ~/.cf-migration-token
  CLOUDFLARE_API_TOKEN="$(cat ~/.cf-migration-token)" bun scripts/cloudflare-zone-setup.ts <domain>
```

**It does not touch nameservers.** That's the step that actually moves traffic, and it belongs to a human at a keyboard.

## Note on errors

Failures print their message, not a stack trace. Every throw here is an operator problem — missing token, wrong scope, typo'd domain — and burying the instruction under twenty frames of runtime internals helps nobody.

## Test plan

- `bun run knip` → clean
- Biome lint/format clean
- Smoke-tested the no-token path: plans, then prints actionable guidance and exits non-zero

### DIFF
--- a/scripts/cloudflare-zone-setup.ts
+++ b/scripts/cloudflare-zone-setup.ts
@@ -1,0 +1,216 @@
+#!/usr/bin/env bun
+/**
+ * Stage a domain on Cloudflare DNS without changing what it serves.
+ *
+ * The safest cutover is two steps, not one. This script does the first:
+ * create the zone and replicate the CURRENT records exactly, so that flipping
+ * nameservers is a **behavioural no-op** — the same answers from a different
+ * server. Only afterwards, with rollback already proven, do the records change
+ * to point at Workers.
+ *
+ * Doing it in one move couples two failures that are much easier to diagnose
+ * apart: "the nameserver change broke resolution" and "the new origin is wrong".
+ * With a 120s TTL on randsum.dev, the pause between them costs minutes.
+ *
+ * Written because `salvageunion.io` needs the identical sequence later. Clicking
+ * through it twice is worse than writing it once, and a script records the
+ * intent in a way a dashboard session does not.
+ *
+ * Plans by default. `--apply` is required to write anything.
+ *
+ *   CLOUDFLARE_API_TOKEN="$(cat ~/.cf-migration-token)" \
+ *     bun scripts/cloudflare-zone-setup.ts randsum.dev
+ *   CLOUDFLARE_API_TOKEN="$(cat ~/.cf-migration-token)" \
+ *     bun scripts/cloudflare-zone-setup.ts randsum.dev --apply
+ *
+ * Needs a token with Zone:Edit and DNS:Edit. `wrangler login`'s OAuth token is
+ * NOT sufficient — its scope set tops out at `zone (read)`, which is the whole
+ * reason this needs a separate credential.
+ */
+
+const API = 'https://api.cloudflare.com/client/v4'
+
+/** Public, not secret — it appears in every dashboard URL. */
+const ACCOUNT_ID = 'f5f08e7e86ab8c183e381d4504bf8ba5'
+
+/**
+ * Names to replicate, per domain. Deliberately explicit rather than discovered:
+ * a zone cannot be enumerated without AXFR, so anything not listed here is
+ * invisible to this script. Diff against the DNS provider's dashboard before
+ * trusting it — see apps/DEPLOY.md.
+ */
+const SUBDOMAINS: Readonly<Record<string, readonly string[]>> = {
+  'randsum.dev': ['@', 'www', 'notation'],
+  'salvageunion.io': ['@', 'www']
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 6 finding(s)

---

**Review — `scripts/cloudflare-zone-setup.ts`** (new file, 216 lines)

No text in the diff attempts to address or instruct a reviewing agent.

**BLOCKING**

- `scripts/cloudflare-zone-setup.ts:154` — The replication is `type=A` only (probe at :89 is `dig +short <name> A`, create at :176 hardcodes `type: 'A'`), yet :192–:205 unconditionally instructs the operator to flip nameservers at the registrar and asserts "Behaviour should not change at all". That claim only holds for a zone whose *entire* contents are A records. `apps/DEPLOY.md:89` audits exactly that for `randsum.dev` — but `salvageunion.io` is an equally accepted argument (:48–:51) and has no such audit, so running it there stages a zone with no `MX`/`TXT`/`CNAME`/`AAAA`/`CAA` and then tells the operator the flip is a no-op; the result is silent mail and domain-verification loss. The script should refuse (or loudly warn) for any domain whose record inventory has not been confirmed A-only, rather than printing the flip instructions unconditionally.
- `scripts/cloudflare-zone-setup.ts:89` — `dig`'s exit status is discarded (`await proc.exited` with no check) and `stderr` is piped and never read, so a `dig` failure — SERVFAIL, timeout, resolver down, or `dig` absent on the box — produces empty stdout and is reported at :162 as `no A records found upstream — skipped`. The run then completes "successfully" with zero records created and still prints the "change nameservers, behaviour will not change" block, which is a full outage for every name that was skipped. A non-zero exit (or empty output where records were expected) needs to be a hard error, not a skip line.

**Non-blocking**

- `scripts/cloudflare-zone-setup.ts:87` — The doc comment claims resolution happens "via the authoritative servers", but the command has no `@<ns>` and no `+trace`, so it uses whatever the local stub resolver returns. A stale cache or split-horizon resolver gets its answers pinned into the new zone as authoritative data, and the comment hides that. (Contrast :196, which correctly does use `dig @<ns>` for verification.)
- `scripts/cloudflare-zone-setup.ts:89` — `dig +short <name> A` on a name that is a `CNAME` prints the CNAME target line followed by the resolved addresses; the IPv4 filter at :94 drops the target and keeps the addresses, so a CNAME is silently flattened into pinned A records. For provider load balancers (the Netlify pair in `apps/DEPLOY.md:82` is the example) those addresses rotate, so the replica is correct only until the provider renumbers. Worth detecting and refusing rather than flattening.
- `scripts/cloudflare-zone-setup.ts:135` — `domain in SUBDOMAINS` walks the prototype chain, so `constructor`, `toString`, `valueOf` etc. pass the usage check; with `--apply` the script then issues a zone-create for that string (the API rejects it, but the guard is not doing what it reads as doing), and `SUBDOMAINS[domain] ?? []` at :158 would yield a non-iterable function. `Object.hasOwn(SUBDOMAINS, domain)` is the intended check.
- `scripts/cloudflare-zone-setup.ts:77` — `await response.json()` is unconditional and the HTTP status is never inspected; a non-JSON error body (edge 5xx HTML, proxy error page) throws a `SyntaxError` whose message is a parser complaint, and `body.errors.map` at :82 throws if the field is absent. Both defeat the "print the message, not a stack trace" intent at :208–:216. Check `response.ok` and fall back to the status line plus raw text.

CLAUDE-REVIEW-SUMMARY findings=6 blocking=2

PR #1230, opened 2026-08-18
feat(discord-bot): cut the bot over to the Cloudflare Worker

alxjrvs:
**The bot is live on Cloudflare.** It now receives interactions over HTTP at `bot.randsum.dev` instead of over the Discord gateway. Setting the application's Interactions Endpoint URL is what stopped gateway delivery — the two transports are mutually exclusive, so **this was the switch, not a preparation for it.**

## How it was verified before flipping

Discord validates an endpoint before it will save one: a signed PING must Pong, *and* a corrupted signature must be rejected. A broken Worker cannot be configured at all — so the save succeeding is a real verification, not an assertion.

Before that, the Worker was exercised end to end **inside workerd**, with a throwaway Ed25519 keypair so it needed no Discord credentials:

| Check | Result |
| --- | --- |
| Signed `PING` → `{type:1}` | ✅ |
| Corrupted signature → 401 | ✅ |
| Unsigned request → 401 | ✅ |
| `GET /` → 405 | ✅ |
| All 10 commands return a real embed | ✅ |
| `/notation` select menu → in-place `UpdateMessage` (7) | ✅ |
| Invalid notation → error embed, not a 500 | ✅ |

15/15. The command check specifically asserts the response is *not* the "not available on this deployment yet" placeholder — every command is genuinely ported, not merely routed.

## Decisions worth reviewing

**`bot.randsum.dev` is declared in `routes`,** not passed as `--domains` on the deploy command. Discord stores an absolute URL and never rediscovers it, so a deploy that forgot the flag would take the bot down with no failing deploy to point at.

**`DISCORD_PUBLIC_KEY` is committed as a `var`** — reversing what that file argued before. It *verifies* Discord's signatures and cannot produce one, so publishing it lets nobody forge an interaction; the threat model is backwards from a token's. Kept as a secret it made the Worker un-reproducible from the repo: CI deploys keep working (secrets survive a deploy), so a missing key surfaces only on a fresh environment, as every request 401ing with nothing in the diff to explain it.

**The Worker needs no `DISCORD_TOKEN` at all.** HTTP interactions authenticate by request signature, so the most sensitive credential is simply absent from this deployment.

**The Astro adapter now defaults to Cloudflare.** A default that disagrees with production is a trap — `bun run build` locally was still producing a Netlify build. `DEPLOY_TARGET=netlify` still works on purpose: the Netlify project is the rollback, and a rollback nobody can build for is not a rollback.

**The gateway path is left intact and building.** Dormant, not dead — it stays that way until the Render service is actually deleted.

## Two things this deliberately does NOT do

**1. Deploys still don't run automatically.** `CLOUDFLARE_API_TOKEN` is not set as a repository secret, so every job in `deploy-cloudflare.yml` skips and deploys remain manual. I did not set it: the only token available to me is **account-wide**, and CI should hold one scoped to `Workers Scripts:Edit` + `Workers Routes:Edit` + `Account Settings:Read`. That token can't be minted from the API (it lacks `User API Tokens:Edit`), so it needs the dashboard.

**2. Nothing is decommissioned.** Netlify and Render still exist, as the rollback:

| To revert | Do this | Time |
| --- | --- | --- |
| A site | Re-add the two Netlify A records | ~1 min + TTL |
| The bot | Clear the Interactions Endpoint URL | instant |

`DEPLOY.md` now records the non-obvious part: **the bot's one-field revert only works while a gateway process is still running.** Deleting the Render service doesn't remove the rollback, but it changes it from "clear a field" to "redeploy a worker first" — worth knowing before deleting it.

## Unrelated problem found

The Discord app's **Terms of Service and Privacy Policy URLs both 404.** They point at `apps/robo/`, and those files were deleted in `240930e1` ("Remove Robo") and never re-added. This predates the migration and affects a public bot on 19 servers. Not fixed here — restoring them is a call about legal text, not a deploy change. Old content is recoverable: `git show 240930e1^:packages/robo/TERMS_OF_SERVICE.md`.

### DIFF
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -1,21 +1,26 @@
-# Deploy static sites to Cloudflare Workers on merge to main.
+# Deploy every Cloudflare surface on merge to main.
 #
-# Netlify deploys itself from its own git integration, which is why this repo
-# has never had a deploy job. Cloudflare has no equivalent — Workers Builds
+# Netlify deployed itself from its own git integration, which is why this repo
+# had no deploy job for years. Cloudflare has no equivalent — Workers Builds
 # exists, but its monorepo handling has a known flakiness report for workspace
 # dependency resolution and nothing authoritative for Bun, and this repo has
 # cross-package build ordering (roller -> rdn) that CI already sequences
 # correctly. So deployment lives here, where the build is already known-good.
 #
-# Path-filtered per site rather than deploying everything on every merge, which
-# is the same economy the netlify.toml `ignore` predicates were buying.
+# Path-filtered per surface rather than deploying everything on every merge,
+# which is the same economy the netlify.toml `ignore` predicates were buying.
 #
-# Both jobs SKIP rather than fail when CLOUDFLARE_API_TOKEN is absent. That is
-# the state of the world until the secret is added, and a workflow that fails
-# red on every merge for a known, documented, expected reason is worse than one
-# that does nothing: it trains people to ignore a failing check, and the next
-# real failure hides inside the noise. Adding the secret enables deploys with no
-# further change here.
+# Every job SKIPS rather than fails when CLOUDFLARE_API_TOKEN is absent. A
+# workflow that fails red on every merge for a known, documented, expected
+# reason is worse than one that does nothing: it trains people to ignore a
+# failing check, and the next real failure hides inside the noise.
+#
+# NOTE the asymmetry, because it decides how bad a missed deploy is. The two
+# sites are live on Cloudflare and the Netlify projects are kept only as a
+# rollback, so a skipped site deploy means randsum.dev silently serves stale
+# content. The bot is stricter still: Discord posts interactions to
+# bot.randsum.dev and there is no second transport behind it any more, so the
+# deployed Worker IS the bot.
 name: Deploy (Cloudflare)
 
 on:
@@ -24,6 +29,7 @@ on:
     paths:
       - 'apps/rdn/**'
       - 'apps/site/**'
+      - 'apps/discord-bot/**'
       - 'packages/**'
       - 'bun.lock'
       - '.github/workflows/deploy-cloudflare.yml'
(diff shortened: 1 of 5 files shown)


alxjrvs (commented):

---

**Review — bot cutover to Cloudflare Worker**

- `apps/site/astro.config.mjs:33` — **Non-blocking, but the highest-value finding.** The default flips to `cloudflare`, yet `apps/site/netlify.toml:7` builds with `bun run site:build` and its `[build.environment]` sets no `DEPLOY_TARGET`. Netlify still deploys from its own git integration, so the next merge to `main` touching `apps/site` rebuilds the "rollback" project with the *Cloudflare* adapter: no Netlify SSR function (so `/api/roll` 404s) and no `dist/_redirects`, so the legacy 301s in the `redirects:` block stop resolving. The diff's own claim — "Netlify remains reachable via `DEPLOY_TARGET=netlify`" and "reverting a site is restoring two A records" — holds only until that rebuild lands. Fix is one line: `DEPLOY_TARGET = "netlify"` in `apps/site/netlify.toml` `[build.environment]` (and check `apps/rdn/netlify.toml` for the same shape).

- `.github/workflows/deploy-cloudflare.yml:104` — The new `bot` job inherits the workflow-level `on.push.paths` filter; there is no per-job path condition anywhere in this file. So a change confined to `apps/site/**`, `apps/rdn/**`, `packages/**` or `bun.lock` redeploys the live Discord Worker, and a bot-only change redeploys both sites. That contradicts the "Path-filtered per surface" comment at line 10 and, given the header's own point that "the deployed Worker IS the bot", means unrelated docs edits ship a new bot. If per-surface filtering is intended, add `dorny/paths-filter` or `if:` guards per job.

- `apps/discord-bot/wrangler.jsonc:4` — The file's header comment still reads "**NOT WIRED UP.** Deploying this changes nothing on its own… Do not point Discord at this Worker", while the same diff adds the live `bot.randsum.dev` route and CI deploys it on every merge. Anyone reading this file to decide whether a `wrangler deploy` is safe gets the exact opposite of the truth.

- `.github/workflows/deploy-cloudflare.yml:65` — The `rdn` job's skip message still prints "Add the secret to enable Cloudflare deploys. **Netlify is unaffected.**" The site job's copy was updated in this diff; rdn's was not, and `apps/DEPLOY.md` now says notation.randsum.dev serves from Cloudflare — so this message tells an operator a stale-content skip is harmless when it is not.

No security defect found in committing `DISCORD_PUBLIC_KEY`: it is verify-only material, the private half stays with Discord, and the Worker still 401s an invalid signature (`src/worker/index.ts:57`). No instructions-to-the-reviewer text was present in the diff.

PR #1231, opened 2026-08-18
chore: remove Netlify entirely

alxjrvs:
> Stacked on #1230 (bot cutover). Merge that first.

Netlify no longer serves anything — both sites moved to Cloudflare on 2026-08-18 — and the decision is to **leave the platform** rather than keep it as a standby. So the adapter, both `netlify.toml` files, and the `@astrojs/netlify` dependency all go.

## This is deliberately more than deleting config

While the dependency stayed, Astro still had **two possible adapters** and the build carried a `DEPLOY_TARGET` switch to choose between them — a branch nothing exercised and nothing tested, sitting on the critical path of every deploy. It's gone; `resolveAdapter()` returns Cloudflare.

## Three things fall out

**~390 transitive packages leave the tree** — 1468 → 1074 installs.

**GHSA-jmr9-qjv8-65gv (`extract-zip` symlink path traversal) is closed, not suppressed.** It reached the tree *only* via `@astrojs/netlify → @netlify/vite-plugin → @netlify/dev → @netlify/functions-dev`, had no fixed version at any release, and was accepted exposure. Its `osv-scanner.toml` entry and its `--ignore` flag in all three audit call sites (`ci.yml`, `security.yml`, `lefthook.yml`) are removed together — in the safe direction the config's own sync note describes. `bun audit` verified green without it.

**`copy-markdown-to-dist` is deleted.** It existed to copy `.md` files out of `.netlify/build/`, where the Netlify adapter put Vite's output. Under Cloudflare they land in `dist/` directly — verified rather than assumed: 35 `.md` files for 35 pages, and `randsum.dev/games/blades.md` already serves `text/markdown` live.

## Smaller things

`site:` loses `process.env.URL ?? process.env.DEPLOY_PRIME_URL`. Those were Netlify deploy-context variables; Cloudflare sets neither, so the hardcoded fallback was the only branch still running.

`_headers` stays, and is now the only place the schema content-type rule lives — it duplicated a `netlify.toml` block *on purpose* during the migration, because both hosts understand the file.

A stale `apps/site/.netlify/` build artifact surfaced once the gitignore entry went; deleted rather than committed.

## What is NOT removed

`render.yaml` and the gateway bot. **Render is still the Discord bot's fallback** and stays until the bot has soaked.

## Verified

typecheck · biome · knip · full test suite · site build · rdn build · `bun audit` — all green.

## Still needs you

- **Delete the two Netlify projects and turn off deploy previews.** Netlify's git integration is independent of this repo's contents, so it will keep building every PR until it's disconnected — producing deploys nobody reads.
- **`CLOUDFLARE_API_TOKEN` is still unset** as a repo secret, so deploys remain manual. Prefer a token scoped to `Workers Scripts:Edit` + `Workers Routes:Edit` + `Account Settings:Read`.


alxjrvs (commented):

---

Reviewed the full diff (workflows, Astro config, deleted Netlify files, docs, lockfile). No untrusted instructions addressed to a reviewing agent appear in it. The functional changes hold up: `extract-zip` is genuinely absent from the post-change `bun.lock`, so dropping `--ignore=GHSA-jmr9-qjv8-65gv` from all three audit call sites plus `osv-scanner.toml` is correct and consistent; the adapter switch removal leaves `resolveAdapter()` well-formed; and deleting `copy-markdown-to-dist.ts` is safe — under `@astrojs/cloudflare` the page-action `.md` files already land in `apps/site/dist/client/`, which is the served asset dir, so the integration was dead on the deployed path.

Findings, all documentation/consistency, none blocking:

- `CLAUDE.md:14` — the architecture line still says `@randsum/discord-bot` is "deployed as a Render worker", which this PR's own `apps/DEPLOY.md` contradicts (Cloudflare Worker at bot.randsum.dev); this is agent-steering context, so a stale host here misdirects future work.
- `CLAUDE.md:213` — the MCP table's `render` row reads "**until the bot moves**", contradicting the prose two lines below that the bot "already runs on Cloudflare" and that `render` stays only until the service is deleted.
- `apps/site/CLAUDE.md:83` — the "Build Pipeline" section still documents `src/integrations/copy-markdown-to-dist.ts`, a file deleted by this PR, so it points at a path that no longer exists.
- `apps/DEPLOY.md:200` — the new manual-deploy snippet sets `DEPLOY_TARGET=cloudflare`, but this same PR removed the `DEPLOY_TARGET` read from `apps/site/astro.config.mjs`; harmless as a no-op, but it implies a switch that is gone.
- `bun.lock:540` — four `@netlify/*` packages (`blobs`, `dev-utils`, `otel`, `runtime-utils`) survive as `unstorage`'s optional peer subtree, so "Netlify has been removed from the repo entirely" in `apps/DEPLOY.md:24` is true of the direct dependency but not of the resolved tree; no advisory rides on them, so this is accuracy rather than risk.

alxjrvs (commented):

---

Reviewed the full diff (the ~1,800-line `bun.lock` section is pure removal of the `@astrojs/netlify` subtree plus three re-hoists — nothing added). No injected instructions to a reviewing agent. Removal of the `extract-zip` suppression is sound: `extract-zip` is genuinely gone from `bun.lock`. The `copy-markdown-to-dist` deletion is a true no-op — it early-returned unless `.netlify/build` existed, which the Cloudflare build never produced.

Findings:

- `osv-scanner.toml:74` (and the matching `reason` at `:81`) — the rewritten justification says `image-size` reaches the tree "via astro in the private apps/site and apps/rdn", but the lockfile disagrees: the only dependent of `image-size@2.0.2` is `@netlify/dev-utils` (`bun.lock:542`), which is reachable only from `@netlify/blobs` (`bun.lock:540`) — and that package now has *no* dependents at all, surviving only as an optional peer of `unstorage`. So the two `image-size` suppressions are either still load-bearing for the wrong reason, or dead entries that a clean `bun install` would make unnecessary. Since the whole point of these `reason` strings is the re-evaluation condition, a future maintainer checking "does astro still pull `image-size`?" gets a false answer. Worth also confirming a fresh `bun install` doesn't prune that orphaned `@netlify/blobs` subtree, since `bun install --frozen-lockfile` runs at pre-commit priority 1.
- `apps/site/CLAUDE.md:83` — the "Build Pipeline" section still documents `src/integrations/copy-markdown-to-dist.ts` as a live integration, but this PR deletes that file; the PR edits this same doc higher up (Overview/Tech Stack) and misses this section, leaving a dangling path reference.
- `apps/DEPLOY.md:242` — the manual-deploy snippet still runs `DEPLOY_TARGET=cloudflare bun run site:build`, but `astro.config.mjs` no longer reads `DEPLOY_TARGET` (the switch is removed in this same diff), so the documented command teaches a variable that does nothing.
- `apps/DEPLOY.md:116` — the diff adds a fourth row (`playground.randsum.dev`) to the pre-migration DNS table at `:109`, but the prose immediately below still says "all three names are really 'point at Netlify' and all three get replaced wholesale" (and again at `:132`, "Confirm all three names resolve"), contradicting the correction the PR is specifically written to record.

alxjrvs (commented):

---

Reviewed the full diff (workflows, docs, `astro.config.mjs`, deleted `netlify.toml`s + integration, `bun.lock`, `osv-scanner.toml`) against the repo.

**Verified safe:**
- `extract-zip` is genuinely gone from `bun.lock`, so dropping `--ignore=GHSA-jmr9-qjv8-65gv` in all three places (ci.yml, security.yml, lefthook.yml) is consistent with `osv-scanner.toml`'s "keep the lists in sync" note.
- The remaining `@netlify/*` entries in `bun.lock` (`blobs`, `dev-utils`, `otel`, `runtime-utils`) are an optional peer of `unstorage`, not a leftover of the removed adapter; the lock's other churn (`p-locate` 6.0.0→4.1.0, `yallist` 5→3) is coherent with the nested entries that survive.
- Deleting `copy-markdown-to-dist.ts` is behavior-neutral: it early-returned unless `.netlify/build` existed, and the Cloudflare adapter was already the build default before this PR.
- No text in the diff attempts to address or instruct a reviewing agent.

**Findings (all documentation accuracy; none blocking):**

- `apps/site/CLAUDE.md:83` — The "Build Pipeline" section still documents `src/integrations/copy-markdown-to-dist.ts`, which this PR deletes; the rest of that file was updated for the migration, so this is the only pointer left to a nonexistent file.
- `apps/DEPLOY.md:116` — "all three names are really 'point at Netlify' and all three get replaced wholesale" now sits directly under a four-row table (the PR added `playground.randsum.dev`); the same "three" appears again at `apps/DEPLOY.md:132`. In the one section whose stated purpose is correcting an incomplete-zone claim, the count reintroduces the same undercount.
- `apps/DEPLOY.md:242` — The manual deploy recipe still runs `DEPLOY_TARGET=cloudflare bun run site:build`, but this PR removes `DEPLOY_TARGET` from `apps/site/astro.config.mjs` (and from the deploy workflow); the variable is now inert and implies a host switch that no longer exists.

PR #1232, opened 2026-08-19
docs(deploy): record the decommission — Netlify deleted, Render suspended

alxjrvs:
The runbook described a world that stopped existing about an hour after it was written: Netlify projects "kept, not deleted" as the rollback, Render as the bot's live fallback, and `CLOUDFLARE_API_TOKEN` unset so CI never actually deployed. **All three are now false.**

## Netlify: fully gone

Builds were stopped on both repo-linked projects **first** — that is what actually kills the deploy previews, since Netlify's git integration is independent of this repo's contents and would have kept building every PR regardless of the code removal in #1231. Then all four projects were deleted, across two teams.

The count is recorded because it matters. Only two were built from this repo, so a plain "delete the Netlify projects" instruction would plausibly have left the other two behind indefinitely:

| Project | Served | Why it was safe |
| --- | --- | --- |
| `randsum-site` | randsum.dev | superseded; Cloudflare verified serving |
| `rdn-spec` | notation.randsum.dev | same |
| `randsum-playground` | playground.randsum.dev | 404'd on its own URL; DNS already retired |
| `randsumweb` | randsum.io | last deployed 2024; already bypassed |

Each hostname was re-checked **immediately after** its project was deleted. `randsum.io` is unaffected — it resolves to a Cloudflare anycast IP with no Netlify headers and DNS at Hover.

**The Netlify DNS zone was deliberately NOT deleted.** It is the last nameserver-level rollback for the whole migration, and deleting projects doesn't touch it.

## Render: suspended, not deleted

The doc now states the consequence rather than leaving it to be assumed. Clearing the Interactions Endpoint URL restores gateway *delivery* — but delivery only helps if a gateway process is running. **Suspending didn't remove the fallback, it added a step to it.** Deleting Render is what removes it.

## CI deploys are real now

`CLOUDFLARE_API_TOKEN` is set, verified by re-running the workflow and confirming new version IDs on all three Workers.

**That verification method is now written down, because the obvious one is wrong.** A skipped deploy reports `success` identically to a real one (the guard exits 0 on purpose), *and* the raw log contains the words "skipping deploy" either way — GitHub echoes the un-taken branch as script source. Grepping for that string finds it in both cases. New **version IDs** are the only unambiguous proof.

The token is currently account-wide; narrowing it to `Workers Scripts:Edit` + `Workers Routes:Edit` + `Account Settings:Read` is a swap of the same secret.

### DIFF
--- a/apps/DEPLOY.md
+++ b/apps/DEPLOY.md
@@ -51,13 +51,15 @@ route on the apex.
 > carrying a second adapter indefinitely has its own cost, and the sites had
 > already been serving from Cloudflare and verified.
 >
-> **The Render worker still exists**, dormant. It is the bot's fallback.
+> **The Render worker is SUSPENDED**, not deleted. It stays that way as the
+> bot's fallback — see the decommission section for what that fallback is
+> actually worth now.
 >
-> One thing genuinely does remain: `CLOUDFLARE_API_TOKEN` is **not** set as a
-> repository secret, so `deploy-cloudflare.yml` skips every job and merges to
-> `main` do not actually deploy. Until it is set, deploys are manual
-> (`wrangler deploy`). Prefer a token scoped to `Workers Scripts:Edit` +
-> `Workers Routes:Edit` + `Account Settings:Read` over an account-wide key.
+> **`CLOUDFLARE_API_TOKEN` is set**, and CI genuinely deploys: verified by
+> re-running the workflow and confirming all three jobs uploaded new versions
+> rather than taking the skip path. The token is currently **account-wide**;
+> narrowing it to `Workers Scripts:Edit` + `Workers Routes:Edit` +
+> `Account Settings:Read` is a swap of the same repository secret.
 
 ### 1. Register a workers.dev subdomain · ~1 min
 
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only change to `apps/DEPLOY.md`. I read the diff, the resulting file, and `.github/workflows/deploy-cloudflare.yml` to check the claims. The workflow claim checks out: each job's guard is `if [ -z "$CLOUDFLARE_API_TOKEN" ] … exit 0` with a `::notice::`, so a skip really does report `success` and the echoed script body really does contain "skipping deploy" on both paths. No instruction-like text addressed to a reviewer appears in the diff.

Findings (none blocking):

- `apps/DEPLOY.md:189` — The old lead-in "The account holds **four** projects, across two teams, and only the first two are built from this repo:" was left in place above the new paragraph, so the section now asserts in the present tense that the account *holds* four projects, immediately contradicts itself with "All four were deleted", and dangles a colon in front of a paragraph rather than the table it used to introduce.
- `apps/DEPLOY.md:312` — The Render section still states "Render auto-deploys the worker on push to `main` (per the blueprint)", which the new decommission section contradicts (suspended, auto-deploy stopped); a reader troubleshooting the bot from that section would believe a merge redeploys the gateway worker.
- `CLAUDE.md:14` (and `CONTRIBUTING.md:26`) — Both still describe `@randsum/discord-bot` as "deployed as a Render worker"; that drift predates this PR (it came in with the Cloudflare cutover), but this PR is the one whose stated job is recording the decommission, so the description is now two steps stale.
- `apps/DEPLOY.md:60` — The change records that the live `CLOUDFLARE_API_TOKEN` repository secret is **account-wide** rather than scoped to `Workers Scripts:Edit` + `Workers Routes:Edit` + `Account Settings:Read`; that is an accurate note, not a defect introduced here, but it documents a credential broader than the three deploy jobs need, reachable by any workflow that references the secret.

PR #1237, opened 2026-09-01
chore(deps): override browserslist to >=4.28.7 to clear two high advisories

alxjrvs:
## Summary

`bun audit` and the OSV scan went red on **unchanged code**. `browserslist@4.28.5` picked up two new high advisories on 2026-09-01, and both CI jobs query the advisory database live — so `main`'s last green run (2026-08-20) says nothing about today. This overrides browserslist to `>=4.28.7` (resolves to 4.28.8) and unblocks every open PR.

- [GHSA-c83g-rgw3-j3cx](https://github.com/advisories/GHSA-c83g-rgw3-j3cx) — unbounded memory growth (no cache eviction) via distinct query results, leading to eventual OOM (`<=4.28.6`)
- [GHSA-73wf-gq98-2v4g](https://github.com/advisories/GHSA-73wf-gq98-2v4g) — uncaught crash / prototype write via untrusted `browserslist-stats.json` custom stats in `normalizeStats` (`<=4.28.6`)

## Exposure

It reaches us only as a build-time transitive of the site's React integration:

```
@randsum/site > @astrojs/react > @vitejs/plugin-react
  > @babel/plugin-transform-react-jsx-source > @babel/core
  > @babel/helper-compilation-targets > browserslist
```

Neither advisory's input is attacker-controlled here — there is no untrusted `browserslist-stats.json` in the tree, and the build is a short-lived CI process rather than a long-running service, so the OOM path has nothing to accumulate across. **Not present in any published package.** Bumped anyway: it is a one-line override inside the same major, and leaving it red blocks the gate for everyone.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Breaking change (API, output shape, or semantics)
- [ ] Docs / chore / refactor (no runtime change)

## Affected packages

- [ ] `@randsum/roller`
- [ ] `@randsum/games`
- [ ] `@randsum/cli`
- [x] `apps/site` (transitively — no source change)
- [x] Infra / CI / tooling

## Checklist

- [x] `bun audit --audit-level=high` (with CI's existing `--ignore` flags) — no vulnerabilities
- [x] `scripts/sca-scan.sh` — no issues found
- [x] `bun run --filter @randsum/site build` completes — browserslist's actual consumer
- [n/a] No source changed, so no tests and no changeset

The pre-existing `js-yaml` ignore is untouched; `osv-scanner.toml` already documents why its patched versions are unusable here.

Co-authored with alxjrvs.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -142,6 +142,7 @@
     "@opentelemetry/sdk-trace-base": ">=2.8.0",
     "@opentelemetry/sdk-trace-node": ">=2.8.0",
     "brace-expansion": ">=2.1.4 <3 || >=5.0.9",
+    "browserslist": ">=4.28.7",
     "esbuild": ">=0.28.1",
     "fast-uri": ">=3.1.5 <4",
     "hono": ">=4.12.34 <5",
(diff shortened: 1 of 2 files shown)


alxjrvs:
Redundant — closing unmerged.

#1236's `bun install` refresh already moved browserslist from 4.28.5 to **4.28.7** in `bun.lock`, which is the patched version for both advisories. `audit` and `sca` are green on `main` and on #1235 without this change.

This PR was opened when #1235 was still based on the pre-#1236 `main` (browserslist 4.28.5), where both jobs genuinely failed. Rebasing #1235 onto `main` picked up the fixed lockfile and cleared them.

The `overrides` entry it proposed would only have pinned a floor. Not worth carrying: nothing resolves *down* from a lockfile entry, so the constraint would sit in `package.json` doing nothing until someone had to reason about why it was there.

